PR #154, opened 2026-01-28
🐛 Fix hourglass refresh indicator on all dashboards

clubanderson:
## Summary
- Add `useRefreshIndicator` hook that guarantees the "Updating" hourglass indicator is visible for at least 500ms on every refresh click, even when cached data returns instantly
- Apply the hook to all 23 dashboard components (Dashboard, Network, Events, Deploy, Compute, Storage, Security, Compliance, Data Compliance, Cost, Alerts, Operators, GitOps, Deployments, Nodes, Logs, Helm, Services, Clusters, Workloads, Pods, Custom Dashboard, Namespaces)
- Move Dashboard.tsx hourglass from the right side to the left side (next to title) for consistency with all other dashboards

## Problem
The "Updating" hourglass indicator was not appearing on Dashboard, Network, Events, and Deploy pages when the user clicked refresh. This was because:
1. Cached data could return so fast that React never rendered the `isRefreshing=true` state
2. The `fetchInProgress` deduplication guard in `fullFetchClusters()` could cause `refetch()` to return silently
3. Dashboard had the hourglass on the wrong side (right) compared to all other dashboards (left)

## Solution
Created a `useRefreshIndicator` hook that manages its own local state, independent of the data hooks:
- `triggerRefresh()` immediately sets `showIndicator=true`, calls the underlying refetch, and uses `setTimeout` to guarantee minimum visibility
- Components use `isRefreshing || showIndicator` for the hourglass condition
- Components use `triggerRefresh` instead of `refetch` for the onClick handler

## Test plan
- [x] Build passes (`npm run build`)
- [x] Verified via Chrome DevTools Protocol on all dashboards:
  - Dashboard ✅, Network ✅, Events ✅, Deploy ✅
  - Compute ✅, Storage ✅, Security ✅, Security Posture ✅
  - Data Compliance ✅, GitOps ✅, Alerts ✅, Clusters ✅
  - Workloads ✅
- [x] Hourglass shows on LEFT side next to title on all dashboards (consistent)

### DIFF
--- a/web/src/hooks/useRefreshIndicator.ts
+++ b/web/src/hooks/useRefreshIndicator.ts
@@ -1,4 +1,4 @@
-import { useState, useCallback, useRef } from 'react'
+import { useState, useCallback, useRef, useEffect } from 'react'
 
 // Minimum time to show the "Updating" hourglass indicator.
 // Ensures the hourglass is always visible even if data returns instantly.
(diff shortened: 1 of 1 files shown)


clubanderson:
Closing — all changes from this PR were already merged via #155 (shared DashboardHeader component) and #156 (event count fix).

PR #1186, opened 2026-02-19
✨: add Flatcar Container Linux monitoring card

aaradhychinche-alt:
### Summary
This PR adds a new **Flatcar Container Linux Status** dashboard card to the KubeStellar Console.  
The card provides visibility into Flatcar-based node OS versions and update status, helping operators monitor node OS health at a glance.

### Key Features
- New `flatcar_status` card following existing console card patterns
- Demo mode support with realistic mock Flatcar node data
- Live mode derives data from Kubernetes Node information
- Displays Flatcar node count, OS version distribution, and update status
- Health status reflects whether all Flatcar nodes are up to date
- Graceful handling of loading, error, empty, and partial-data states

### Implementation Details
- Card UI implemented using shared console components
- Data logic encapsulated in a dedicated hook
- Demo data separated for clarity and maintainability
- Card registered with default grid size `{ w: 6, h: 4 }` under the Provisioning category

### Verification
- [x] Console runs locally via `./startup-oauth.sh`
- [x] Card appears in **Add Cards** catalog
- [x] Demo mode displays mock Flatcar data correctly
- [x] Live mode derives Flatcar data from Kubernetes Nodes
- [x] Card renders correctly with no console or TypeScript errors

### Related Issue
kubestellar/console-marketplace#59

### DIFF
--- a/pkg/k8s/client.go
+++ b/pkg/k8s/client.go
@@ -1268,7 +1268,7 @@ func (m *MultiClusterClient) GetClusterHealth(ctx context.Context, contextName s
 		health.ErrorType = classifyError(errMsg)
 		health.ErrorMessage = errMsg
 		health.Issues = append(health.Issues, fmt.Sprintf("Failed to list nodes: %v", nodesErr))
-	} else {
+	} else if nodes != nil {
 		health.NodeCount = len(nodes.Items)
 		var totalCPU int64
 		var totalMemory int64
@@ -1302,7 +1302,7 @@ func (m *MultiClusterClient) GetClusterHealth(ctx context.Context, contextName s
 	}
 
 	// Process pods - non-fatal, fall back to cached values on timeout
-	if podsErr == nil {
+	if podsErr == nil && pods != nil {
 		health.PodCount = len(pods.Items)
 		var totalCPURequests int64
 		var totalMemoryRequests int64
@@ -1335,7 +1335,7 @@ func (m *MultiClusterClient) GetClusterHealth(ctx context.Context, contextName s
 	}
 
 	// Process PVCs - non-fatal, fall back to cached values on timeout
-	if pvcsErr == nil {
+	if pvcsErr == nil && pvcs != nil {
 		health.PVCCount = len(pvcs.Items)
 		for _, pvc := range pvcs.Items {
 			if pvc.Status.Phase == corev1.ClaimBound {


clubanderson:
## Code Review — Flatcar Container Linux Monitoring Card

### Overall
Good concept for monitoring Flatcar nodes. Several issues need attention, one high-priority:

### Issues

#### 1. Missing isDemoData wiring (HIGH)
The component uses `useCachedData` but does **not** destructure `isDemoData` or pass it to `useCardLoadingState()`. This means when demo/fallback data is shown, the card will NOT display the Demo badge or yellow outline — a regression against the established pattern.

**Fix:** Destructure `isDemoData` from the hook return and pass it to `useCardLoadingState()`:
```ts
const { data, isDemoData } = useCachedFlatcarStatus();
// ...
useCardLoadingState({ isLoading, error, isDemoData });
```

#### 2. Missing `consecutiveFailures` forwarding (Medium)
The custom hook tracks `consecutiveFailures` but doesn't expose it to the component. If the card should show degraded state after repeated failures, this needs to be forwarded.

#### 3. Hardcoded English error string (Medium)
Error display uses a hardcoded English string instead of a `t()` call. Wrap with i18n.

#### 4. Demo data health contradicts logic (Low)
The demo data marks nodes as `"healthy"` but some have `lastUpdate` timestamps that would be considered "outdated" by the live logic. This could confuse users inspecting demo mode — demo data should be internally consistent.

#### 5. Raw `/api/v1/nodes` fetch may not be proxied (Medium)
The hook fetches `/api/v1/nodes` directly. Verify this goes through the console's backend proxy. If the frontend makes this call directly to the K8s API, it will fail in production (CORS, auth). The existing pattern uses the backend API proxy.

#### 6. Version sort doesn't handle "unknown" (Low)
If a node's Flatcar version can't be determined, it may be set to `"unknown"`. The version sorting/comparison logic should handle this gracefully.

#### 7. Spurious whitespace changes (Low)
A few files have whitespace-only changes unrelated to the Flatcar card. Consider reverting those.

### What looks good
- Card registration pattern is correct
- Node grouping by version is a nice UX touch
- Preset config is clean

clubanderson:
Hey! Thanks for contributing this card — the UI work looks great and we appreciate the effort. 🙏

However, this card was merged with **demo data only** and no real backend wiring, which means it never shows actual data from live clusters. We've had to temporarily disable it in PR #2211 until it's properly connected to real cluster data.

**To get this card re-enabled, it needs:**

1. **A Go backend handler** — Query the actual Kubernetes resources from all clusters using `DeduplicatedClusters()` (see `pkg/api/handlers/crds.go` for an example)
2. **A frontend hook with API-first + demo fallback** — Try the real API endpoint first, fall back to demo data with `isDemoData: true` flag (see `web/src/hooks/useCRDs.ts`)
3. **Proper `isDemoData` wiring** — So the card shows a "Demo" badge when displaying synthetic data via `useCardLoadingState()`

PR #2210 (CRD Health) is the reference implementation for this pattern.

We'd love for you to complete this work and submit a follow-up PR! The card UI is already done — it just needs the data layer. If you have questions about the backend wiring pattern, feel free to ask. We're happy to help.

**Going forward, all new cards must include real backend data + demo fallback to qualify for approval and merge.**

PR #2300, opened 2026-03-12
🐛 Fix storageKey mismatch in Deployments dashboard

### LINKED ISSUE #2238: storageKey mismatch between Deployments.tsx and deployments.ts config causes card layout resets
## User Request

**Type:** bug
**Submitted by:** @Arpit529Srivastava
**Console Request ID:** 2ea33fba-b327-475d-9c69-82c1512f18d1

## Description

What happened: Deployments.tsx persists card layout under 'kubestellar-deployments-cards' but web/src/config/dashboards/deployments.ts declares storageKey: 'deployments-dashboard-cards'. These are two different localStorage keys, so layout saved by one path is invisible to the other.

What I expected: A single consistent storageKey used everywhere for the deployments dashboard.

Steps to reproduce:

Customize card layout on the Deployments dashboard
Navigate away and back
Layout resets because the config and page use different keys

---

clubanderson:
## Summary
- Change `DEPLOYMENTS_CARDS_KEY` from `'kubestellar-deployments-cards'` to `'deployments-dashboard-cards'` to match the config in `deployments.ts`
- This was causing card layout customizations to be lost across navigations

## Test plan
- [ ] Customize card layout on deployments page, navigate away and back — layout should persist
- [ ] Check localStorage — should only have `deployments-dashboard-cards` key, not the old one

Fixes #2238

### DIFF
--- a/web/src/components/deployments/Deployments.tsx
+++ b/web/src/components/deployments/Deployments.tsx
@@ -7,10 +7,16 @@ import { useDrillDownActions } from '../../hooks/useDrillDown'
 import { useUniversalStats, createMergedStatValueGetter } from '../../hooks/useUniversalStats'
 import { StatBlockValue } from '../ui/StatsOverview'
 import { DashboardPage } from '../../lib/dashboards/DashboardPage'
-import { getDefaultCards } from '../../config/dashboards'
+import { getDefaultCards, deploymentsDashboardConfig } from '../../config/dashboards'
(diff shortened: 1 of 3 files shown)


clubanderson:
@Arpit529Srivastava — (correcting earlier tag) you filed the original issue for this fix. Could you test with the Netlify Deploy Preview once it's ready and let us know if it looks good? Thanks!

PR #2312, opened 2026-03-12
🐛 Fix auto-update polling not restarting on toggle

### LINKED ISSUE #2245: Bug : Auto Update Not Working
## User Request

**Type:** bug
**Submitted by:** @Abhishek-Punhani
**Console Request ID:** f46b7175-ec2f-42d1-b244-d7e76e398e0d

## Description

Auto-update notifier shows a new commit, but the agent never downloads or applies it because the UpdateChecker’s background loop stays disabled unless the user previously enabled auto-update, and even

---

clubanderson:
## Summary
- Fixes #2245
- The polling `useEffect` in `useVersionCheck` was missing `autoUpdateEnabled` in its dependency array, so toggling auto-update on/off didn't restart/stop the polling interval
- Added `autoUpdateEnabled` to the dependency array and added an early-return guard so the effect correctly skips polling when disabled

## Test plan
- [ ] Toggle auto-update off → verify polling stops (no network requests in DevTools)
- [ ] Toggle auto-update on → verify polling resumes at the configured interval
- [ ] Check Netlify Deploy Preview

### DIFF
--- a/web/src/hooks/useVersionCheck.test.tsx
+++ b/web/src/hooks/useVersionCheck.test.tsx
@@ -6,15 +6,20 @@ import type { GitHubRelease, ParsedRelease } from '../types/updates'
 import { UPDATE_STORAGE_KEYS } from '../types/updates'
 
 // ---------------------------------------------------------------------------
-// Mock external dependencies so the hook can mount without a live agent
+// Mock external dependencies so the hook can mount without a live agent.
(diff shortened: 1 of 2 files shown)


clubanderson:
@Abhishek-Punhani — (correcting earlier tag) you filed the original issue for this fix. Could you test with the Netlify Deploy Preview once it's ready and let us know if it looks good? Thanks!

PR #2473, opened 2026-03-13
🐛 Harden dynamic card sandbox against global access and prototype pollution

clubanderson:
## Summary
- Shadow dangerous browser globals (`window`, `document`, `fetch`, `localStorage`, `eval`, `Function`, `XMLHttpRequest`, `WebSocket`, etc.) with `undefined` in the dynamic card `new Function()` scope
- Freeze the scope object with `Object.freeze()` to prevent prototype pollution attacks
- Whitelisted scope entries (React, Lucide icons, card hooks) remain accessible; everything else is blocked

## Security Context
Finding **C2** from penetration test: `new Function()` in `compiler.ts` executes user-defined card code. While the existing whitelisted scope prevents direct DOM/window access, this hardening adds defense-in-depth by:
1. Explicitly shadowing ~30 dangerous globals so card code cannot escape the sandbox even if `globalThis` resolution changes
2. Freezing the scope to prevent `scope.React.__proto__` type attacks

`unsafe-eval` remains in the CSP because `new Function()` requires it. A full Web Worker/iframe sandbox would allow removing it but is a larger architectural change.

## Test plan
- [ ] Verify existing Tier 2 dynamic cards still render correctly
- [ ] Verify card code cannot access `window`, `document`, `fetch`, or `localStorage`
- [ ] Build passes (pre-existing i18n type errors in OpenFeature card are unrelated)

### DIFF
--- a/web/src/lib/dynamic-cards/compiler.ts
+++ b/web/src/lib/dynamic-cards/compiler.ts
@@ -4,6 +4,20 @@ import type { CardComponentProps } from '../../components/cards/cardRegistry'
 import { getDynamicScope } from './scope'
 
 /**
+ * Browser globals that must be shadowed inside the dynamic card sandbox.
+ * Each is bound to `undefined` so card code cannot reach the real objects.
+ */
+const BLOCKED_GLOBALS = [
+  'window', 'document', 'globalThis', 'self', 'top', 'parent', 'frames',
+  'fetch', 'XMLHttpRequest', 'WebSocket', 'EventSource',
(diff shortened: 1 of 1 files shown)


clubanderson:

Copilot code review found **0 code suggestion(s)** and **1 general comment(s)**.

- **`web/src/lib/dynamic-cards/compiler.ts`** (line 56): The comment says "Freeze each scope value that is an object to prevent prototype pollution" but `Object.freeze(scope)` o

---

PR #2534, opened 2026-03-13
🐛 Fix start.sh: atomic binary writes and kc-agent restart

### LINKED ISSUE #2526: bug: start.sh writes to a file that may be in use
### Describe the bug

It appears that https://github.com/kubestellar/console/blob/01726eb914f6d95498038ef5d5cbefe832d181d4/start.sh (that's the latest commit in `main`, as of this writing) is designed to be something that is intended to be runnable concurrently with running `console` and/or `kc-agent` processes, but there is a concurrency problem. `start.sh` writes to `console` and `kc-agent` --- in particular, it may write a new version, and the write may be partial. This is a bad thing to do to an executable file that a running process is based on.

I asked Claude code / Opus 4.6 about `curl -o`, and got the following.

```
❯ Does `curl -o $filename` simply write to the file, or does it first write to a
temporary file and then move that to `$filename`?

⏺ curl -o $filename writes directly to the file — it does not use a temporary file
  and then rename/move it. It opens (or creates) the file at the start of the
  transfer and writes data to it as it arrives.

  This means if the download is interrupted, you'll be left with a partial file at
  that path. It also means another process reading the file during download could
  see incomplete content.

  If you want atomic behavior (write to temp, then move), you'd need to do it
  yourself, e.g.:

  curl -o "$filename.tmp" "$url" && mv "$filename.tmp" "$filename"
```

### Environment

Any

### Steps To Reproduce

Discussed above

### Expected Behavior

No trashing running processes.

```shell

```

### Want to contribute?

### Additional Context

_No response_

### LINKED ISSUE #2530: bug: start.sh does not restart kc-agent
### Describe the bug

start.sh downloads both `console` and `kc-agent`, and restarts the `console` process if there is one, but does not restart the `kc-agent` process if there is one. This seems problematic, there could be relevant changes to `kc-agent`.

### Environment

Any

### Steps To Reproduce

Discussed above

### Expected Behavior

Installing a new version fully installs the new version.

```shell

```

### Want to contribute?

### Additional Context

_No response_

clubanderson:
## Summary
- Download binaries to temp files then atomically `mv` into place to prevent partial writes corrupting running processes
- Restart `kc-agent` process when updating, matching the existing `console` restart behavior

Fixes #2530
Fixes #2526

## Test plan
- [ ] Run start.sh with console and kc-agent already running
- [ ] Verify both binaries are updated and both processes restarted
- [ ] Verify partial download failure doesn't corrupt existing binaries

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -125,18 +125,29 @@ resolve_version() {
 }
 
 # --- Download and extract ---
+# Downloads to a temp file then atomically moves into place to prevent
+# partial writes from corrupting a running binary.
 download_binary() {
     local name="$1" version="$2" platform="$3"
     local url="https://github.com/${REPO}/releases/download/${version}/${name}_${version#v}_${platform}.tar.gz"
+    local tmp_extract_dir
+    tmp_extract_dir=$(mktemp -d)
 
     echo "  Downloading ${name} ${version} (${platform})..."
     if ! curl -sSL --fail -o "/tmp/${name}.tar.gz" "$url" 2>/dev/null; then
         echo "  Warning: Failed to download ${name} from ${url}"
+        rm -rf "$tmp_extract_dir"
         return 1
     fi
 
-    tar xzf "/tmp/${name}.tar.gz" -C "$INSTALL_DIR"
+    # Extract to a temporary directory first
+    tar xzf "/tmp/${name}.tar.gz" -C "$tmp_extract_dir"
     rm -f "/tmp/${name}.tar.gz"
+
+    # Atomically move the binary into the install directory
+    chmod +x "$tmp_extract_dir/${name}" 2>/dev/null || true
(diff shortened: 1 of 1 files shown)


clubanderson:

Copilot code review found **0 code suggestion(s)** and **4 general comment(s)**.

- **`start.sh`** (line 141): The tarball is still downloaded to a fixed path (`/tmp/${name}.tar.gz`). Concurrent runs of `start.sh` can clobber each 
- **`start.sh`** (line 270): The kc-agent restart path always sends SIGKILL after a fixed 2s sleep, even if the process already exited after SIGTERM.
- **`start.sh`** (line 150): `tmp_extract_dir` is created via `mktemp -d` (typically under /tmp), then `mv` moves the extracted binary into `$INSTALL
- **`start.sh`** (line 149): Because the script uses `set -e`, failures in `tar` extraction or `mv` will exit the whole script immediately and skip c

---

PR #2535, opened 2026-03-13
🐛 Prevent SQLite data loss during Helm upgrades

clubanderson:
## Summary
- **Add `helm.sh/resource-policy: keep`** to the PVC template so Helm never deletes the SQLite volume during upgrades — this directly caused GPU reservation data loss today
- **Add optional CronJob** (`backup.enabled=true`) that copies `console.db` to a separate backup PVC every 6 hours, retaining the 20 most recent backups
- **Add init container auto-restore** — if the main DB is empty/missing on startup (e.g., after PVC recreation), automatically restores from the latest backup
- **Immediate fix applied**: annotated the live PVC on vllm-d with `helm.sh/resource-policy: keep`

All backup features are opt-in (`backup.enabled` defaults to `false`). The `keep` annotation is always applied.

## Test plan
- [ ] `helm template` with `backup.enabled=false` (default) — no CronJob, no init container, no backup PVC
- [ ] `helm template` with `backup.enabled=true` — CronJob, init container, backup PVC all render correctly
- [ ] Both PVCs have `helm.sh/resource-policy: keep` annotation
- [ ] Deploy to test cluster, verify backup CronJob creates DB copies
- [ ] Delete main DB file, restart pod — verify init container restores from backup

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build stage - Backend
-FROM golang:1.24-alpine AS backend-builder
+FROM golang:1.25-alpine AS backend-builder
 
 WORKDIR /app
 


clubanderson:

- **`deploy/helm/kubestellar-console/templates/deployment.yaml`** (line 35): The restore initContainer/backup volume blocks are gated only by `backup.enabled` + `backup.autoRestore`, but they alway
- **`deploy/helm/kubestellar-console/templates/deployment.yaml`** (line 38): The initContainer does not set a `securityContext`, so it won’t inherit the chart’s container-level hardening (`runAsNon
- **`deploy/helm/kubestellar-console/templates/cronjob-db-backup.yaml`** (line 44): The backup job claims to take a consistent SQLite snapshot via `.backup`, but it actually uses `cp` on the live DB file.

---

PR #2565, opened 2026-03-14
⚡ Race CRD/deployment checks in parallel for instant Trestle detection

clubanderson:
## Summary
- All 6 detection checks (3 CRDs + 3 deployments) now run in parallel via `Promise.allSettled` instead of sequentially
- Non-installed clusters resolve in ~3s instead of up to 30s (6 sequential × 5s timeout)
- Reduced CRD check timeout from 5s to 3s — missing CRDs fail near-instantly on reachable clusters

## Before
Sequential: CRD1 → CRD2 → CRD3 → Dep1 → Dep2 → Dep3 = up to 30s per cluster

## After
Parallel: all 6 checks race simultaneously = ~3s per cluster

## Test plan
- [x] Build passes
- [x] TypeScript clean
- [ ] Trestle card loads quickly on compliance dashboard
- [ ] Card shows "Checking clusters... X/11" counter that increments rapidly

### DIFF
--- a/web/src/hooks/useTrestle.ts
+++ b/web/src/hooks/useTrestle.ts
@@ -26,8 +26,8 @@ const REFRESH_INTERVAL_MS = 120_000
 /** Cache TTL: 2 minutes — matches refresh interval */
 const CACHE_TTL_MS = 120_000
 
-/** Timeout for CRD existence check */
-const CRD_CHECK_TIMEOUT_MS = 5_000
+/** Timeout for CRD/deployment existence check (fast — missing resources fail instantly) */
+const CRD_CHECK_TIMEOUT_MS = 3_000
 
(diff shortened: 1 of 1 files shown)


clubanderson:

- **`web/src/hooks/useTrestle.ts`** (line 214): The new detection logic still awaits `Promise.allSettled(allChecks)`, so it cannot actually short-circuit on the “first 
- **`web/src/hooks/useTrestle.ts`** (line 209): This change always schedules all 6 kubectl checks even if the first CRD lookup succeeds, which increases total kubectl t
- **`web/src/hooks/useTrestle.ts`** (line 30): `CRD_CHECK_TIMEOUT_MS` is now used for both CRD and deployment checks, so the constant name is misleading. Consider rena

---

PR #2772, opened 2026-03-16
🐛 Fix alert notifications: dedup, deep link, and repeated firing

clubanderson:
## Summary

Fixes three alert notification bugs reported by user:

### 1. Repeated macOS notifications (no dedup)

**Before:** Every 60-second evaluation cycle sent a new browser notification as long as the condition was true. Disk pressure → notification every minute.

**After:** `notifiedAlertKeysRef` tracks which alerts already triggered a notification. A notification fires once when the condition first appears. The key is cleared when the alert resolves, so if the condition recurs after resolution, a fresh notification fires.

### 2. Disk pressure notification opens wrong page

**Before:** Clicking the notification opened `?card=cluster_health` — the cluster health card, where disk pressure isn't visible.

**After:** Opens a node drilldown (`?drilldown=node&cluster=X&node=Y&issue=DiskPressure`) showing the affected node's conditions, events, and kubectl access. Falls back to cluster drilldown if the node name can't be parsed.

### 3. Missing affected node in alert details

**Before:** Alert details only had `clusterName` and `issue` — no way to know which node had disk pressure.

**After:** Parses node name from the issue string and includes `affectedNode` in the alert details.

### Evaluators fixed

| Evaluator | Notification dedup | Deep link fix |
|-----------|-------------------|---------------|
| disk_pressure | Added | Fixed (node drilldown) |
| gpu_health_cronjob | Added | Already correct |
| nightly_e2e_failure | Added | Already correct |
| memory_pressure | N/A (no notification) | N/A |

## Test plan

- [ ] Build passes
- [ ] Trigger disk pressure alert → notification fires once
- [ ] Wait 60s → no repeat notification
- [ ] Click notification → opens node drilldown (not cluster health)
- [ ] Resolve disk pressure → notification key cleared
- [ ] Disk pressure recurs → new notification fires

### DIFF
--- a/web/src/contexts/AlertsContext.tsx
+++ b/web/src/contexts/AlertsContext.tsx
@@ -165,6 +165,11 @@ export function AlertsProvider({ children }: { children: ReactNode }) {
   const rulesRef = useRef(rules)
   rulesRef.current = rules
 
+  // Track which alert dedup keys have already triggered a browser notification.
+  // Prevents the same alert from sending repeated macOS notifications on every
+  // evaluation cycle. Keys are cleared when the alert resolves.
+  const notifiedAlertKeysRef = useRef<Set<string>>(new Set())
+
   // CronJob health results cache — fetched async, read synchronously by evaluator
   const cronJobResultsRef = useRef<Record<string, GPUHealthCheckResult[]>>({})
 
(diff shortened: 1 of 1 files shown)


clubanderson:

- **`web/src/contexts/AlertsContext.tsx`** (line 881): `diskPressureIssue` strings coming from the backend are formatted like `"DiskPressure on %d node(s): %s"` (see pkg/k8s/c
- **`web/src/contexts/AlertsContext.tsx`** (line 170): The comment says notification dedup keys are cleared “when the alert resolves”, but `resolveAlert()` only updates `alert
- **`web/src/contexts/AlertsContext.tsx`** (line 1046): `notifiedAlertKeysRef` is used to dedup nightly E2E browser notifications, but the key you add (`${rule.id}::${guide.acr

---

PR #2846, opened 2026-03-17
🐛 fix: set isDone on SSE error to prevent benchmark retry loop

### LINKED ISSUE #2816: 🐛 Latest Benchmark Card Always Shows "Demo" Badge & Mock Data When Local Agent is Running
## User Request

**Type:** bug
**Submitted by:** @Abhishek-Punhani
**Console Request ID:** 7be5941c-06f3-4427-8529-3636af095893

## Summary

The **Latest Benchmark** card displays a `Demo` badge and renders synthetic/mock data even when a local `llm-d` agent is actively running. Users expect live benchmark metrics from their agent, but the card silently falls through to `generateBenchmarkReports()` mock data without any actionable error or explanation.

### The Demo Fallback Chain

The issue is caused by a **multi-layer fallback chain** that treats a missing `GOOGLE_DRIVE_API_KEY` environment variable as equivalent to having no live data — and then falls back silently to mock data with the `Demo` badge:

#### Layer 1 — Backend returns 503 when `GOOGLE_DRIVE_API_KEY` is not set

```go
// pkg/api/handlers/benchmarks.go:488-492
if h.apiKey == "" {
    return c.Status(503).JSON(fiber.Map{
        "error":  "benchmark data not configured — set GOOGLE_DRIVE_API_KEY",
        "source": "unavailable",
    })
}
```

The backend benchmark API is **exclusively backed by Google Drive**. When the env var is missing (common in local dev), both `/api/benchmarks/reports` and `/api/benchmarks/reports/stream` return HTTP 503.

#### Layer 2 — The SSE singleton starts immediately and receives an error

```ts
// web/src/hooks/useBenchmarkData.ts:108-186
fetch(`/api/benchmarks/reports/stream?since=...`, ...)
  .then(async (res) => {
    if (!res.ok || !res.body) {
      // 503 → sets error, isStreaming=false, reports=[]
      streamState = { ...streamState, isStreaming: false, error: `Stream error: ${res.status}` }
      notifySubscribers()
      return
    }
    ...
  })
```

The module-level SSE singleton fires immediately on first subscriber. When it gets 503, it sets `reports: []` and `isStreaming: false` but **never reaches `isDone: true`** — it just silently stops.

#### Layer 3 — `useCache` fallback fires with `demoWhenEmpty: true`

```ts
// web/src/hooks/useBenchmarkData.ts:196-219
const cacheResult = useCache<BenchmarkReport[]>({
  key: 'benchmark-reports',
  demoData: DEMO_REPORTS,        // ← pre-generated mock reports
  demoWhenEmpty: true,           // ← activates demo when live data is empty
  fetcher: async () => {
    if (stream.reports.length > 0 && stream.isDone) return stream.reports
    // Fallback: non-streaming endpoint — also returns 503
    const res = await fetch(`/api/benchmarks/reports?...`)
    if (res.status === 503) throw new Error('BENCHMARK_UNAVAILABLE')
    ...
  },
})
```

The `useCache` hook is configured with `demoWhenEmpty: true`. Since `state.data` is `[]` (empty after the SSE 503 + REST 503), the hook's `shouldFallbackToDemo` condition is `true`:

```ts
// web/src/lib/cache/index.ts:931-932
const shouldFallbackToDemo = effectiveEnabled && demoWhenEmpty && stableDemoData !== undefined
  && !state.isLoading && Array.isArray(state.data) && (state.data as unknown[]).length === 0
```

This sets `isDemoFallback: true`.

#### Layer 4 — `BenchmarkHero` renders mock data and reports `isDemoData: true`

```tsx
// web/src/components/cards/llmd/BenchmarkHero.tsx:95-97
const { data: liveReports, isDemoFallback, ... } = useCachedBenchmarkReports()
const effectiveReports = useMemo(
  () => isDemoFallback ? generateBenchmarkReports() : (liveReports ?? []),
  [isDemoFallback, liveReports]
)
useReportCardDataState({ isDemoData: isDemoFallback, ... })
```

When `isDemoFallback` is `true`, `effectiveReports` is overridden with fresh mock data and the card's `Demo` badge is activated — even though a local agent is running perfectly.

---

## The Core Design Problem

> The benchmark card's data pipeline is **exclusively tied to Google Drive** (`GOOGLE_DRIVE_API_KEY`). There is **no mechanism to serve locally-available benchmark data** from a running llm-d agent. When Google Drive is not configured (typical in local dev), the only data path is mock/demo data.

This creates a **misleading user experience**: the card implies there's no connected agent, when in reality the agent is running fine — just without Google Drive credentials.

### Secondary Problem — SSE Singleton Never Reaches `isDone`

When the SSE stream returns a 503, `streamState.isDone` stays `false`. The `useCachedBenchmarkReports` hook checks `stream.isDone` before delegating to the `useCache` REST fallback:

```ts
// useBenchmarkData.ts:204-207
if (stream.reports.length > 0 && stream.isDone) {
  return stream.reports
}
```

Since `isDone` is never set to `true`, the hook **always falls through** to the REST endpoint on every cache refresh interval — causing repeated 503 calls in a loop.

---

## Steps to Reproduce

1. Run the console backend **without** setting `GOOGLE_DRIVE_API_KEY`
2. Have a local `llm-d` agent running
3. Open the dashboard and observe the **Latest Benchmark** card
4. Notice the `Demo` badge is shown; data is mock (Llama-3.2-1B-Instruct → H200, etc.)
5. Open DevTools → Network → confirm `/api/benchmarks/reports/stream` returns 503

---

## Expected Behavior

- When a local llm-d agent is running, the card should show live benchmark metrics (or at least a meaningful "Not connected to Google Drive — configure `GOOGLE_DRIVE_API_KEY` to enable benchmark history" message)
- The `Demo` badge should only appear when **no** real data source is available **AND** the user is not connected to a live agent
- The SSE singleton should set `isDone: true` on error to prevent repeated 503 retry loops

---

## Actual Behavior

- Card shows `Demo` badge with fully synthetic metrics (hardcoded model/GPU combinations)
- No visible error is surfaced; the card looks completely functional
- DevTools shows repeated 503 calls to `/api/benchmarks/reports` every hour (cache refresh interval)

---

## Affected Files

| File | Issue |
|---|---|
| [`useBenchmarkData.ts`](file:///home/manav/dev_ws/kubestellar/console/web/src/hooks/useBenchmarkData.ts) | SSE error handling never sets `isDone: true`; `demoWhenEmpty: true` triggers silently on 503 |
| [`BenchmarkHero.tsx`](file:///home/manav/dev_ws/kubestellar/console/web/src/components/cards/llmd/BenchmarkHero.tsx) | Renders mock data + `Demo` badge when `isDemoFallback` is true with no error shown to user |
| [`benchmarks.go`](file:///home/manav/dev_ws/kubestellar/console/pkg/api/handlers/benchmarks.go) | No fallback to local agent data; entire benchmark pipeline requires Google Drive API key |
| [`benchmarkMockData.ts`](file:///home/manav/dev_ws/kubestellar/console/web/src/lib/llmd/benchmarkMockData.ts) | `DEMO_REPORTS` is generated at module load; used as fallback without notice |

---

---

clubanderson:
## Summary
- Set `isDone: true` when SSE stream returns error (503/non-ok), not just `isStreaming: false`
- Also set `isDone: true` on SSE `error` event and network-level fetch errors
- Prevents repeated 503 retry loop against benchmark REST endpoint
- The `useCache` fetcher now correctly sees the stream is done and won't loop

Fixes #2816

### DIFF
--- a/web/src/hooks/useBenchmarkData.ts
+++ b/web/src/hooks/useBenchmarkData.ts
@@ -111,7 +111,7 @@ function startGlobalStream(since: string) {
   })
     .then(async (res) => {
       if (!res.ok || !res.body) {
-        streamState = { ...streamState, isStreaming: false, error: `Stream error: ${res.status}` }
+        streamState = { ...streamState, isStreaming: false, isDone: true, error: `Stream error: ${res.status}` }
         notifySubscribers()
         return
       }
@@ -165,7 +165,7 @@ function startGlobalStream(since: string) {
                 streamState = { ...streamState, isDone: true, isStreaming: false, status: 'done' }
                 notifySubscribers()
               } else if (eventType === 'error') {
-                streamState = { ...streamState, error: rawData, isStreaming: false, status: 'error' }
+                streamState = { ...streamState, error: rawData, isStreaming: false, isDone: true, status: 'error' }
                 notifySubscribers()
               }
             }
@@ -180,7 +180,7 @@ function startGlobalStream(since: string) {
     })
     .catch((err) => {
       if (err.name !== 'AbortError') {
-        streamState = { ...streamState, error: err.message, isStreaming: false }
+        streamState = { ...streamState, error: err.message, isStreaming: false, isDone: true }
         notifySubscribers()
       }
     })


clubanderson:

- **`web/src/hooks/useBenchmarkData.ts`** (line 115): When the SSE response is non-OK / missing a body you set `isDone: true`, but `status` is left as whatever it was previou
- **`web/src/hooks/useBenchmarkData.ts`** (line 183): In the network-level `.catch` path you now set `isDone: true` but don’t update `status` (so it can remain `'connecting'`

---

PR #2934, opened 2026-03-19
🐛 fix: deduplicate browser notifications with 5-minute cooldown

clubanderson:
## Summary
- Replace `Set<string>` with `Map<string, timestamp>` for notification dedup tracking
- Enforce 5-minute cooldown between repeat notifications for the same alert
- Keys no longer deleted on resolve — prevents re-notification when clusters flap

## Root Cause
The dedup key was deleted when a cluster alert resolved. If the cluster briefly appeared reachable (transient network), the key was deleted, then on the next 30-second check it was unreachable again and the notification fired again. This caused dozens of duplicate macOS notifications.

## Before/After
- **Before**: Notification every 30s for each unreachable cluster
- **After**: One notification, then 5-minute cooldown before any repeat

### DIFF
--- a/web/src/contexts/AlertsContext.tsx
+++ b/web/src/contexts/AlertsContext.tsx
@@ -166,9 +166,13 @@ export function AlertsProvider({ children }: { children: ReactNode }) {
   rulesRef.current = rules
 
   // Track which alert dedup keys have already triggered a browser notification.
-  // Prevents the same alert from sending repeated macOS notifications on every
-  // evaluation cycle. Keys are cleared when the alert resolves.
-  const notifiedAlertKeysRef = useRef<Set<string>>(new Set())
+  // Maps dedup key → timestamp (ms) of last notification. Prevents the same alert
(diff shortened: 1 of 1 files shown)


clubanderson:

Copilot code review found **0 code suggestion(s)** and **4 general comment(s)**.

- **`web/src/contexts/AlertsContext.tsx`** (line 175): NOTIFICATION_COOLDOWN_MS is defined inside the component but captured by multiple useCallback evaluators; this will typi
- **`web/src/contexts/AlertsContext.tsx`** (line 176): `notifiedAlertKeysRef` now retains keys indefinitely (no deletes on resolve), so this Map can grow without bound over a 
- **`web/src/contexts/AlertsContext.tsx`** (line 830): The cooldown check is duplicated in multiple evaluators and currently does multiple Map lookups and multiple `Date.now()
- **`web/src/contexts/AlertsContext.tsx`** (line 917): This comment is now inaccurate: it says the notification dedup key is cleared when DiskPressure resolves, but the delete

---

PR #3164, opened 2026-03-22
fix: normalize nvidia-operators JSON key and add dashboard import rollback

### LINKED ISSUE #3138: Inconsistent JSON Schema in /api/mcp/nvidia-operators
## User Request

**Type:** bug
**Submitted by:** @Arpit529Srivastava
**Console Request ID:** fe177ec6-9d4a-4fca-95a5-6a239b39b0e6

## Description

What happened: The endpoint returns differing root dictionary keys ("operator" vs "operators") depending entirely on whether the cluster query parameter is provided. This irregularity forces front-end components to type-template or defensively parse schema variations. It often causes silent component crashes when UI iterators run into an Object map instead of an Array.

What I expected: The API should always return an array under the "operators" root key for structural consistency, even if it only contains a single element when queried down to a singular host cluster.

Steps to reproduce:

Run curl -H "Authorization: Bearer <TOKEN>" http://localhost:8080/api/mcp/nvidia-operators (Returns array: {"operators": [...]})
Run curl -H "Authorization: Bearer <TOKEN>" http://localhost:8080/api/mcp/nvidia-operators?cluster=minikube (Returns object: {"operator": {...}})

---

### LINKED ISSUE #3145: Partial Database Import Failure (No Transaction Rollback) - /api/dashboards/import
## User Request

**Type:** bug
**Submitted by:** @Arpit529Srivastava
**Console Request ID:** 59229ecf-c346-426c-8f5e-8793f1d77135

## Description

What happened: When importing a Dashboard via JSON, the ImportDashboard
 API sequentially creates the Dashboard and then each of its associated Cards using store.CreateCard(). If any card creation fails natively (e.g. database constraint or connection drop), the handler immediately returns a 500 error. However, it does not execute a SQL transaction rollback for the items it already created.

How it affects users: This heavily impacts users who routinely share or backup their dashboard configurations. If an import fails midway, they are left with a corrupted, half-built dashboard in their UI that they must manually delete, causing confusion and cluttering the database with orphaned records.

What I expected: Bulk imports should be wrapped in an atomic SQL transaction so that any card creation failure automatically rolls back the entire dashboard creation, completely preventing corrupted partial states.

Steps to reproduce:

Fire a POST /api/dashboards/import with a valid dashboard JSON but append a malformed CardExport configuration.
Observe a 500 Internal Server Error return.
Refresh the frontend UI and notice the partially created dashboard permanently sitting in the navigation bar.

---

clubanderson:
## Summary
- **#3138**: Single-cluster `/api/mcp/nvidia-operators?cluster=X` now returns `"operators"` (array) instead of `"operator"` (singular object), matching the multi-cluster response shape and preventing frontend parse mismatches.
- **#3145**: `ImportDashboard` handler now rolls back the partially-created dashboard when a card creation fails mid-import, preventing orphaned records in the store.

## Test plan
- [ ] `go build ./cmd/console` passes
- [ ] Single-cluster nvidia-operators response uses `"operators"` key with array value
- [ ] Multi-cluster nvidia-operators response unchanged
- [ ] Dashboard import with invalid card data cleans up on failure

Closes #3138
Closes #3145

### DIFF
--- a/pkg/api/handlers/dashboard.go
+++ b/pkg/api/handlers/dashboard.go
@@ -254,6 +254,8 @@ func (h *DashboardHandler) ImportDashboard(c *fiber.Ctx) error {
 			Position:    ce.Position,
 		}
 		if err := h.store.CreateCard(card); err != nil {
+			// Rollback: delete the partially-created dashboard and any cards
+			_ = h.store.DeleteDashboard(dashboard.ID)
 			return fiber.NewError(fiber.StatusInternalServerError, "Failed to create card")
 		}
 	}
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -657,7 +657,7 @@ func (h *MCPHandlers) GetNVIDIAOperatorStatus(c *fiber.Ctx) error {
 			log.Printf("internal error: %v", err)
 			return c.Status(500).JSON(fiber.Map{"error": "internal server error"})
 		}
-		return c.JSON(fiber.Map{"operator": status, "source": "k8s"})
+		return c.JSON(fiber.Map{"operators": []*k8s.NVIDIAOperatorStatus{status}, "source": "k8s"})
 	}
 
 	return c.Status(503).JSON(fiber.Map{"error": "No cluster access available"})


clubanderson:

- **`pkg/api/handlers/dashboard.go`** (line 260): Rollback currently ignores the result of DeleteDashboard. If the delete fails (e.g., transient DB issue), the handler wi
- **`pkg/api/handlers/dashboard.go`** (line 260): ImportDashboard gained rollback behavior on CreateCard failure, but there is no test asserting that the partially-create
- **`pkg/api/handlers/mcp.go`** (line 660): The single-cluster NVIDIA operators endpoint now changes its JSON schema to return an array under "operators". There isn

---

PR #3309, opened 2026-03-23
fix: ClearAlert panic on variable-length device type suffixes

### LINKED ISSUE #3289: DeviceTracker.ClearAlert : Panics on Malformed Alert IDs via Incorrect String Slicing
## User Request

**Type:** bug
**Submitted by:** @arnavgogia20
**Console Request ID:** 51e62f72-374b-475f-a457-0fe6b1f90622

## Problem

`DeviceTracker.ClearAlert()` uses hard-coded string slicing (`alertID[:len(alertID)-len("/gpu")-1]`) to extract the node key from an alert ID. This assumes all alert IDs end with a 3-character device type suffix like `"/gpu"` or `"/nic"`. However, the system generates alert IDs with variable-length suffixes such as `"/nvme"` (4 chars), `"/infiniband"` (10 chars), `"/gpu-driver"` (10 chars), `"/mofed-driver"` (12 chars), and `"/spectrum-scale"` (15 chars). This causes incorrect key extraction, corrupted max-count resets, and potential out-of-bounds panics when the suffix length exceeds the hard-coded offset.

---

## Root Cause

`device_tracker.go:473`

```go
parts := alertID[:len(alertID)-len("/gpu")-1] // rough extraction
```

This line assumes `len("/gpu") + 1 = 5` chars should be stripped from the end. For an alert ID like `"cluster/node/spectrum-scale"`, this produces `"cluster/node/spectrum-s"` — a non-existent key. For very short alert IDs, the subtraction could produce a negative index, causing a panic.

The subsequent `switch` statement on lines `479-485` only handles `"gpu"`, `"nic"`, and `"nvme"` — missing `"infiniband"`, `"sriov"`, `"rdma"`, `"mellanox"`, `"mofed-driver"`, `"gpu-driver"`, and `"spectrum-scale"`.

---

## Reproduction Steps

1. Deploy a cluster with Mellanox HCAs or NVIDIA GPUs with driver monitoring labels  
2. Wait for `DeviceTracker` to detect a driver state change (e.g., MOFED goes not-ready)  
3. Call `ClearAlert("cluster/nodeName/mofed-driver")` via the `/devices/alerts/clear` HTTP endpoint  
4. The function computes an incorrect node key, fails to reset max counts, causing the same alert to immediately re-fire on the next scan cycle  

---

## Expected Behavior

`ClearAlert` should correctly parse the alert ID regardless of device type suffix length, properly extract the node key, and reset the correct device counter in `maxCounts`.

---

## Suggested Fix

Replace the brittle string slicing with `strings.LastIndex`:

```go
func (t *DeviceTracker) ClearAlert(alertID string) bool {
    t.mu.Lock()
    defer t.mu.Unlock()
    if _, ok := t.alerts[alertID]; ok {
        delete(t.alerts, alertID)
        // Extract node key: everything before the last "/"
        lastSlash := strings.LastIndex(alertID, "/")
        if lastSlash < 0 { return true }
        nodeKey := alertID[:lastSlash]
        deviceType := alertID[lastSlash+1:]
        if counts, ok := t.maxCounts[nodeKey]; ok {
            if history, ok := t.history[nodeKey]; ok && len(history) > 0 {
                latest := history[len(history)-1].Counts
                switch deviceType {
                case "gpu":            counts.GPUCount = latest.GPUCount
                case "nic":            counts.NICCount = latest.NICCount
                case "nvme":           counts.NVMECount = latest.NVMECount
                case "infiniband":     counts.InfiniBandCount = latest.InfiniBandCount
                case "sriov":          counts.SRIOVCapable = latest.SRIOVCapable
                case "rdma":           counts.RDMAAvailable = latest.RDMAAvailable
                case "mellanox":       counts.MellanoxPresent = latest.MellanoxPresent
                case "mofed-driver":   counts.MOFEDReady = latest.MOFEDReady
                case "gpu-driver":     counts.GPUDriverReady = latest.GPUDriverReady
                case "spectrum-scale": counts.SpectrumScale = latest.SpectrumScale
                }
                t.maxCounts[nodeKey] = counts
            }
        }
        return true
    }
    return false
}
```

---

## Impact

**Reliability:** Device alerts for non-GPU types (InfiniBand, NVMe, MOFED drivers, etc.) cannot be properly cleared. Once triggered, these alerts persist indefinitely even after the device issue is resolved, polluting the monitoring UI and causing **alert fatigue**.

In edge cases with short alert IDs, this causes a **runtime panic**, crashing the agent process.

---

clubanderson:
Closes #3289

## Summary

- Replace hard-coded string slicing (`alertID[:len(alertID)-len("/gpu")-1]`) with `strings.LastIndex(alertID, "/")` to correctly extract the node key and device type from alert IDs of any length
- Add a guard for alert IDs that contain no slash to prevent negative index panics
- Extend the `switch` statement in `ClearAlert` to handle all device types that `processSnapshot` generates alerts for: `infiniband`, `sriov`, `rdma`, `mellanox`, `mofed-driver`, `gpu-driver`, and `spectrum-scale` (previously only `gpu`, `nic`, and `nvme` were handled)

## Root cause

`ClearAlert()` assumed all device type suffixes were 3 characters long (like "gpu" or "nic"), stripping exactly `len("/gpu")+1 = 5` characters from the end of the alert ID. For longer suffixes like "spectrum-scale" (15 chars) or "mofed-driver" (12 chars), this produced a corrupted node key that wouldn't match any entry in `maxCounts`, so the baseline was never reset and the alert would immediately re-fire on the next scan. For very short alert IDs, the subtraction could produce a negative index, causing a runtime panic.

## Test plan

- [x] `go build ./...` passes
- [ ] Verify `ClearAlert("cluster/node/gpu")` still works correctly (existing behavior)
- [ ] Verify `ClearAlert("cluster/node/mofed-driver")` correctly resets `MOFEDReady` baseline
- [ ] Verify `ClearAlert("cluster/node/spectrum-scale")` correctly resets `SpectrumScale` baseline
- [ ] Verify `ClearAlert("malformed-no-slash")` returns `true` without panicking

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/pkg/agent/device_tracker.go
+++ b/pkg/agent/device_tracker.go
@@ -492,20 +492,39 @@ func (t *DeviceTracker) ClearAlert(alertID string) bool {
 		delete(t.alerts, alertID)
 		// Also reset the max count to current to prevent re-alerting
 		// Extract node key from alert ID (format: "cluster/node/deviceType")
-		parts := alertID[:len(alertID)-len("/gpu")-1] // rough extraction
-		if counts, ok := t.maxCounts[parts]; ok {
+		lastSlash := strings.LastIndex(alertID, "/")
+		if lastSlash < 0 {
+			return true
+		}
+		nodeKey := alertID[:lastSlash]
+		deviceType := alertID[lastSlash+1:]
+		if counts, ok := t.maxCounts[nodeKey]; ok {
 			// Get current counts from latest history
-			if history, ok := t.history[parts]; ok && len(history) > 0 {
+			if history, ok := t.history[nodeKey]; ok && len(history) > 0 {
 				latest := history[len(history)-1].Counts
-				switch {
-				case alertID[len(alertID)-3:] == "gpu":
+				switch deviceType {
+				case "gpu":
 					counts.GPUCount = latest.GPUCount
-				case alertID[len(alertID)-3:] == "nic":
+				case "nic":
 					counts.NICCount = latest.NICCount
-				case alertID[len(alertID)-4:] == "nvme":
+				case "nvme":
 					counts.NVMECount = latest.NVMECount
+				case "infiniband":
+					counts.InfiniBandCount = latest.InfiniBandCount
+				case "sriov":
+					counts.SRIOVCapable = latest.SRIOVCapable
+				case "rdma":
+					counts.RDMAAvailable = latest.RDMAAvailable
+				case "mellanox":
+					counts.MellanoxPresent = latest.MellanoxPresent
+				case "mofed-driver":
+					counts.MOFEDReady = latest.MOFEDReady
+				case "gpu-driver":
+					counts.GPUDriverReady = latest.GPUDriverReady
+				case "spectrum-scale":
+					counts.SpectrumScale = latest.SpectrumScale
 				}
-				t.maxCounts[parts] = counts
+				t.maxCounts[nodeKey] = counts
 			}
 		}
 		return true


clubanderson:

- **`pkg/agent/device_tracker.go`** (line 526): ClearAlert now has a growing deviceType switch that must be kept in sync with the deviceType strings used in processSnap
- **`pkg/agent/device_tracker.go`** (line 501): This change fixes a panic and parsing bug in ClearAlert and adds support for multiple deviceType suffixes, but there are

---

PR #3811, opened 2026-03-28
fix: prevent HardwareHealthCard from overriding user's explicit view choice

### LINKED ISSUE #3798: bug report : HardwareHealthCard auto-switches to alerts view whenever activeAlertCount > 0 — overrides user's explicit view choice
## User Request

**Type:** bug
**Submitted by:** @shivansh-source
**Console Request ID:** 0df5a0a5-66a3-44ee-8290-c3b36ac38738

## Description

web/src/components/cards/HardwareHealthCard.tsx
If a user switches to the Inventory view to check hardware specs, any data refresh that returns alerts will forcibly switch them back to the Alerts view. The user's explicit navigation choice is silently overridden on every data refresh.

---

clubanderson:
Closes #3798

## Summary
- Added a `userSelectedView` ref that tracks whether the user has explicitly clicked a view tab (Inventory or Alerts)
- The auto-switch `useEffect` now only fires on initial load (before any user interaction), not on subsequent data refreshes
- All three user-triggered `setViewMode` calls now set `userSelectedView.current = true` to suppress future auto-switching

## Test plan
- [ ] Open dashboard with active hardware alerts -- card should auto-switch to Alerts view on initial load
- [ ] Click "Inventory" tab -- view switches to Inventory
- [ ] Wait for data refresh (or trigger manual refresh) -- view stays on Inventory, not forced back to Alerts
- [ ] Reload the page -- auto-switch to Alerts should work again on fresh load

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/web/src/components/cards/HardwareHealthCard.tsx
+++ b/web/src/components/cards/HardwareHealthCard.tsx
@@ -95,6 +95,10 @@ export function HardwareHealthCard() {
   const lastUpdate = hwData.lastUpdate ? new Date(hwData.lastUpdate) : null
 
   const [viewMode, setViewMode] = useState<ViewMode>('inventory')
+  // Track whether the user has explicitly chosen a view tab.
+  // When true, auto-switch logic is suppressed so data refreshes
+  // don't override the user's choice.
+  const userSelectedView = useRef(false)
   const [showSnoozed, setShowSnoozed] = useState(false)
   const [snoozeMenuOpen, setSnoozeMenuOpen] = useState<string | null>(null)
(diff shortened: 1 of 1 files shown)


clubanderson:

Copilot code review found **0 code suggestion(s)** and **1 general comment(s)**.

- **`web/src/components/cards/HardwareHealthCard.tsx`** (line 267): The comment says the auto-switch runs only on “initial load”, but the effect will still run on every subsequent `activeA

---

PR #4427, opened 2026-04-03
Fix remaining TS errors in tsconfig.json and vite.config.ts

clubanderson:
## Summary

Fixes the last 4 VS Code TypeScript errors:
- `tsconfig.json`: Silence `baseUrl` deprecation warning with `ignoreDeprecations: "5.0"`
- `vite.config.ts:66`: Add `string` type to `html` parameter in `transformIndexHtml`
- `vite.config.ts:71-72`: Use `algorithms: ['gzip']` array instead of deprecated `algorithm: 'gzip'` string (vite-plugin-compression2 API change)

## Test plan
- [ ] `npm run build` passes
- [ ] Gzip + brotli compressed assets still generated in dist/
- [ ] VS Code shows 0 errors in tsconfig.json and vite.config.ts

### DIFF
--- a/web/tsconfig.json
+++ b/web/tsconfig.json
@@ -16,6 +16,7 @@
     "noUnusedParameters": true,
     "noFallthroughCasesInSwitch": true,
     "noUncheckedSideEffectImports": true,
+    "ignoreDeprecations": "5.0",
     "baseUrl": ".",
     "paths": {
       "@/*": ["src/*"]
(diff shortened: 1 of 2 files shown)


web/tsconfig.json
aaradhychinche-alt, line 1: i think we need to use "ignoreDeprecations": "6.0", to solve the problem !!

PR #5134, opened 2026-04-07
⚠️ Revert memoization removal — cards crash with infinite re-renders

clubanderson:
## CRITICAL — Production broken

Reverts PR #5126. Removing useMemo/useCallback caused React error #185 (Maximum update depth exceeded) across ALL cards on console.kubestellar.io.

Root cause: Callbacks recreated on every render trigger useEffect dependency loops in child components. React Compiler's auto-memoization doesn't cover hooks returning callbacks used as useEffect deps in consumers.

## Test plan
- [x] Cards render without crashes after revert


clubanderson:

- **`web/src/components/compliance/Compliance.tsx`** (line 56): `filteredClusterNames` is created as a new `Set` on every render, so the `realData` `useMemo` will also recompute every 
- **`web/src/components/mission-control/ClusterAssignmentPanel.tsx`** (line 67): `healthyClusters` (and `removedClusters`) are derived with `.filter(...)` outside of `useMemo`, so they get a new array 
- **`web/src/components/gpu/GPUReservations.tsx`** (line 325): `calendarWeeks` is wrapped in `useMemo`, but it depends on `getReservationDayRange`, which is currently re-created on ev

---

Issue #6149, opened 2026-04-10
perf: dashboard navigation is slow — 461 React commits per navigation


clubanderson:
## Reported by

@aarushrai (Slack #console — 2026-04-10 10:40 AM)

> @everyone do we feel like navigating between dashboards is a bit more slower now?
> I don't know if its just me?
> I used react dev tools profiler to get a flamegraph:
> theres 461 commits so maybe thats the problem?

## Smoking gun

A single dashboard-to-dashboard navigation produces **461 React commits** in the DevTools profiler. The flamegraph shows deeply nested Context Providers wrapping every `SortableCard` / `CardWrapper` instance, and a long chain of `Suspense` / `RouteProvider` / `DashboardProvider` / `CardDataContext` calls under each one.

A correct navigation should be ~1-5 commits (route swap + initial render + a few suspended chunks resolving). 461 commits means every state change in some upstream provider is forcing a tree-wide re-render — most likely:

1. **A context provider whose value object is recreated on every render** (no `useMemo`), so every consumer re-renders even when the underlying data is unchanged
2. **Cluster filter or theme state plumbed through a context that fires N times during route mount**
3. **A `useEffect` that calls `setState` synchronously on mount, looping until React bails out**
4. **The card-grid layout calculation re-running on every parent re-render** (not memoized) and triggering another setState

## Investigation tasks

- [ ] Reproduce in dev mode and capture a fresh profiler trace at \`http://localhost:8080\` after a clean reload
- [ ] Identify which Context Providers are firing the most. Top suspects:
  - \`CardDataContext\` (every card subscribes)
  - \`DashboardContext\` / route-level state
  - \`useTheme\` provider
  - \`useDemoMode\` provider
  - \`useGlobalFilters\` (cluster filter)
- [ ] Check whether any provider's \`value=\` prop is a fresh object/array literal each render — that's the canonical cause of context cascades
- [ ] Verify card components are wrapped in \`React.memo\` and that their props don't include unstable references
- [ ] Check the \`useCardCollapse\` subscriber set added in PR #6096 — it's per-cardId so it shouldn't fan out, but worth verifying it doesn't notify all subscribers on a single change
- [ ] Check if dnd-kit / SortableContext from PR #6096 is firing internal updates on hover/drag-detect

## Nightly regression test

@clubanderson asked to add a React DevTools profiler nightly test:

> i will add a react dev tools profiler nightly test

The test should:
1. Open a dashboard via Playwright in dev mode (or a special profiling build)
[...]

PR #6152, opened 2026-04-10
test(e2e): React commit budget test runs every 2h with auto-issue regression reporting

clubanderson:
## Summary

Adds a Playwright perf test that counts React commits during a
dashboard-to-dashboard navigation and fails if they exceed the budget
(default 50). This is the regression guard for #6149, where a single
navigation produced hundreds of commits.

## Cadence: every 2 hours (not nightly)

- `cron: '15 */2 * * *'` — runs 12 times/day, ~30s of runner time per run
  (~360 runner-minutes/month, well under the free-tier budget)
- **Why 2h vs nightly:** when a regression lands, the bisect window is
  ~2h of merges instead of ~24h. On a repo that merges 10-20 PRs/day
  that's the difference between 1-2 suspects and 10-20.
- **Concurrency:** `concurrency: { group: perf-react-commits,
  cancel-in-progress: true }` — a slow run can't pile up behind the next
  tick.

## Auto-filed regression issues with dedup

On test failure, a follow-up `report-failure` job:

1. Downloads `web/perf-result.json` (written by the test in
   `test.afterEach`, so it's captured on pass and fail)
2. Searches for an existing open issue whose title starts with
   `[perf-regression] dashboard navigation commits`
3. **If found:** comments on it with the new run details (dedup — no
   issue spam)
4. **If not found:** creates a new issue with labels
   `kind/bug,priority/high,triage/accepted`

Both the comment and the new-issue body include:

- Commit count observed vs budget
- Link to the failing Actions run
- HEAD commit sha + short message
- A list of merges in the bisect window (since the last successful run
  of this same workflow — falls back to last 20 merges if there is no
  prior success)

Permissions scoped to exactly what's needed: `contents: read`, `issues:
write`, `actions: read`.

## Files

- `.github/workflows/perf-react-commits.yml` — renamed from
  `-nightly.yml`, rewritten for 2h cadence + report-failure job
- `web/e2e/perf/react-commits.spec.ts` — added `test.afterEach` that
  writes `perf-result.json` with `{commits, budget, navigatedTo,
  timestamp, runId}`
- `web/package.json` — `test:e2e:perf:commits` script (from previous
  commit)

## Acceptance

- Once #6149 lands, the next scheduled run goes green and stays green.
- The budget is set high enough (50) that legitimate reactivity
  (suspense, hydration) doesn't trip it — but low enough to catch
  cascades like #6149 which produced hundreds of commits.

## Test plan

- [x] `npm run build` passes locally
- [x] `npx tsc --noEmit` passes for the test file
- [x] Workflow YAML validates
- [ ] First scheduled run (or manual `workflow_dispatch`) completes
- [ ] Intentionally break the budget to verify the report-failure job
      creates an issue, then verify a second failure comments instead
      of filing a duplicate


clubanderson:
Closing in favor of #6160 (generalized perf scaffold). The scaffold covers everything this PR does (React commits every 2h with auto-issue dedup) plus bundle size and Lighthouse TTFI signals using one shared notifier script and one reusable workflow. Single-PR superset of this work.

PR #7559, opened 2026-04-13
✨ feat(ci): auth login smoke test — 30-min cron + nightly

clubanderson:
## Problem

Commit `25e464fa` (#6593) removed the `token` field from `/auth/refresh` as a "security" measure. This broke OAuth login for 24 hours. The existing startup smoke test, contract tests, and nightly dashboard-health workflow all passed — none of them tested the actual auth flow.

## Fix

Two new test surfaces, **neither in the PR CI path** (zero slowdown for contributors):

### 1. `auth-login-smoke.yml` — every 30 minutes
- Checks production (`console.kubestellar.io`) reachability and OAuth redirect
- Starts a local dev-mode backend and verifies the `/auth/refresh` contract:
  - `token` field present and non-empty (the field that was removed)
  - `onboarded` field present
- On failure: creates an issue labeled `auth-smoke-failure` + `priority/critical`
- Deduplicates — won't create a second issue if one is already open

### 2. `nightly-dashboard-health.yml` — new `auth-contract` job
- Same `/auth/refresh` contract check, runs alongside the existing 30+ dashboard route tests
- Shares the failure → issue creation path already in the nightly

## What this would have caught

If this had existed on Apr 11:
- The 30-min cron would have fired within 30 min of the merge
- The `/auth/refresh` contract check would have failed (`token` field missing)
- An issue would have been auto-created with `priority/critical`
- Total time-to-detect: **≤30 minutes** instead of **~24 hours**

## Test plan
- [ ] Build passes (no source changes, only workflow files)
- [ ] After merge: trigger `auth-login-smoke.yml` manually and verify it passes
- [ ] Verify nightly picks up the new `auth-contract` job on next run

### DIFF
--- a/.github/workflows/auth-login-smoke.yml
+++ b/.github/workflows/auth-login-smoke.yml
@@ -191,5 +191,5 @@ jobs:
               repo: context.repo.repo,
               title,
               body,
-              labels: ['auth-smoke-failure', 'priority/critical'],
+              labels: ['auth-smoke-failure', 'priority/critical', 'ai-needs-human'],
             });
--- a/.github/workflows/auth-login-smoke.yml
+++ b/.github/workflows/auth-login-smoke.yml
@@ -1,0 +1,195 @@
+name: Auth Login Smoke Test
+
+# Verifies the OAuth login flow is functional on console.kubestellar.io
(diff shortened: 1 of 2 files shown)


clubanderson:
This is an existing PR (#7559), not a bug to fix. Skipping.

PR #7640, opened 2026-04-13
fix: non-blocking cluster data fetching for instant dashboard rendering

clubanderson:
## Summary

- Adds `onSettled` callback to `settledWithConcurrency` so partial results are pushed to the UI immediately as each cluster responds, instead of waiting for all clusters (including unreachable ones with 15s timeouts) to complete
- Converts the sequential `for...of` loop in the events progressive fetcher to concurrent `settledWithConcurrency` with per-task progress
- Updates 8 multi-cluster fetch functions (`fetchFromAllClusters`, `fetchPodIssuesViaAgent`, `fetchDeploymentsViaAgent`, `fetchWorkloadsFromAgent`, `fetchSecurityIssuesViaKubectl`, `fetchLLMdServers`, `fetchLLMdModels`, ISO27001 audit) to use the new `onSettled` callback

Reachable clusters now render in <100ms while unreachable clusters time out in the background without blocking dashboard switches.

Fixes dashboard 5+ second blocking on "My Clusters" and "Cluster Admin" switches when unreachable clusters are present.

## Test plan

- [ ] Switch between dashboards with a mix of reachable and unreachable clusters — reachable data should appear instantly
- [ ] Verify cards show progressive data (partial results render as clusters respond)
- [ ] Verify `npm run build` passes
- [ ] Verify demo mode still works correctly on console.kubestellar.io

### DIFF
--- a/web/src/hooks/useCachedData.ts
+++ b/web/src/hooks/useCachedData.ts
@@ -161,17 +161,29 @@ async function fetchFromAllClusters<T>(
     return addClusterField ? items.map(item => ({ ...item, cluster })) : items
   })
 
-  const settled = await settledWithConcurrency(tasks)
-
-  // Aggregate fulfilled results and count failures
+  // Use onSettled callback to push partial results to the UI as each
+  // cluster responds, instead of waiting for all clusters (including
+  // unreachable ones with long timeouts) to complete.
   const accumulated: T[] = []
   let failedCount = 0
-  for (const result of settled) {
+
+  const settled = await settledWithConcurrency(tasks, undefined, (result) => {
     if (result.status === 'fulfilled') {
       accumulated.push(...result.value)
       onProgress?.([...accumulated])
     } else {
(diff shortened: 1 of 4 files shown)


clubanderson:
Review: Good approach — progressive rendering via onSettled removes the waterfall block on unreachable clusters. One bug to fix:

**fetchFromAllClusters double-counts failedCount**

The onSettled callback increments failedCount for each rejection. Then the 'Final aggregation pass' (if accumulated.length === 0) re-iterates settled and increments failedCount again. When all clusters fail (say 5), failedCount ends up as 10, but settled.length is 5 — so the condition failedCount === settled.length is false and the throw is silently skipped. This breaks the agent fallback path.

Fix: either reset failedCount = 0 before the final loop, or (cleaner) remove the final aggregation pass entirely — the onSettled callback already handles everything.

Rest of the changes look clean and the pattern is applied consistently.

PR #7854, opened 2026-04-13
✨ feat(rewards): HMAC-signed attribution footer — stop github.com issue gaming

clubanderson:
## Problem

Users were opening issues directly on github.com with a `bug` label to collect the 300-point reward intended for UX-friction-tested console submissions. The rewards classifier had no way to tell where an issue came from.

## Fix

Backend mints an HMAC-signed footer when creating issues via `/api/feedback/submit`. Rewards classifier verifies the signature before awarding 300/100 pts; issues without valid attribution score as 50 pts.

| Attack | Before | After |
|---|---|---|
| Open issue via github.com with `bug` label | 300 pts | 50 pts |
| Forge fake footer on github.com | 300 pts | 50 pts (HMAC fails) |
| Copy valid footer from console issue to new github.com issue | 300 pts | 50 pts (title binding) |
| Reuse same title + footer | 300+300 | 300 + 50 (nonce replay) |

## Required action before merge

Set **`CONSOLE_ATTRIBUTION_SECRET`** as a repo secret — 32+ bytes of random hex. The `.env` on prod needs it too. If unset, the system fails safe (all issues score as web-UI, so rewards drop but never inflate).

## Test plan
- [x] `go test ./pkg/api/handlers/... -run Attribution` — 12 cases pass
- [x] `go build ./...` clean
- [ ] After merge: set `CONSOLE_ATTRIBUTION_SECRET`, submit a test issue via console, verify it gets 300 pts
- [ ] Open a test issue on github.com with bug label, verify it gets 50 pts

### DIFF
--- a/pkg/api/handlers/attribution.go
+++ b/pkg/api/handlers/attribution.go
@@ -1,0 +1,210 @@
+// Attribution signing for console-submitted issues.
+//
+// Problem: the rewards leaderboard awards 300 points for bug reports, but
+// we only want to award that for issues submitted through the console UI
+// — not for issues opened directly on github.com. Without a signature,
+// contributors can open issues on github.com with a "bug" label and get
+// the 300-point reward reserved for UX-friction-tested console submissions.
+//
+// Solution: when the console backend creates an issue on behalf of the
+// user, append an HMAC-signed attribution footer to the issue body. The
+// rewards classifier verifies the signature before awarding console-tier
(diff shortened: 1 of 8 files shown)


clubanderson:
Superseded by GitHub App approach — see follow-up PR. Moving from HMAC-signed footer (secret on backend) to GitHub App attestation (performed_via_github_app set by GitHub itself, no shared secret to forge).

Issue #7865, opened 2026-04-14
perf: split clusterCache into data + UI-indicator caches to unblock SPA navigation


clubanderson:
## Problem

Navigating away from the **main Dashboard** (`/`) or **My Clusters** (`/my-clusters`) requires more than one click. The URL updates (history.push fires) but `<Outlet>` doesn't commit the new route for ~1s+. Other dashboards (Deploy with 16 cards, Security, Insights, etc.) navigate instantly on a single click.

## Root cause

Both pages are heavy consumers of `useClusters()`, which subscribes to a single shared `clusterCache`. That cache mixes two kinds of state:

- **Data state**: `clusters[]`, `lastUpdated`, `consecutiveFailures`
- **UI indicator state**: `isRefreshing`, `isLoading`, `lastRefresh`, `error`

Every poll tick (`CLUSTER_POLL_INTERVAL_MS`), every per-cluster health-check completion (`checkHealthProgressively` fans these out in parallel), and every WebSocket kubeconfig change calls `updateClusterCache()` → `notifyClusterSubscribers()` → synchronous `setLocalState(cache)` on every component using the hook. On Dashboard/MyClusters that's many subscribers re-rendering as urgent setStates many times per second, starving React Router v7's internal `startTransition` for navigation.

Verified via Chrome CDP: at URL `/deploy`, `<main>` still contained `data-testid="dashboard-page"` even after multiple seconds and a second click.

## Why prior attempts failed (PR #7833)

- **Per-subscriber wrap** of `setLocalState` in `startTransition` (one consumer): didn't help — many other components subscribe, all still firing urgent updates.
- **Notify-source wrap** of `notifyClusterSubscribers()` in `startTransition` (all consumers): unblocked navigation (CDP showed Deploy committing in ~988ms vs never), but caused a regression — the `isRefreshing: true → false` transition at the end of each fetch got perpetually superseded by the next fetch's "start" update, so the refresh spinner on the logo never stopped.

## Proposed fix

Split `clusterCache` into two separate cache objects with separate subscriber sets:

- **`clusterDataCache`**: `{ clusters[], lastUpdated, consecutiveFailures, isFailed }` — notify wrapped in `startTransition` (interruptible, can yield to navigation)
- **`clusterUICache`**: `{ isRefreshing, isLoading, lastRefresh, error }` — notify stays urgent (so spinner state always commits)

Update `useClusters()` to subscribe to both and merge into a single return value for backward compatibility with consumers.

## Acceptance

[...]

PR #7876, opened 2026-04-14
fix: throttle cluster cache save + dedupe ClusterAdmin counts

clubanderson:
## Summary

Two fixes from investigating issue #7865 (\"needs 2 clicks to navigate away from Dashboard / My Clusters\"):

### 1. Throttle `saveClusterCacheToStorage`

Per-cluster health checks call `updateSingleClusterInCache` → `saveClusterCacheToStorage` on every completion. With 24+ kubeconfig contexts (12 unique clusters × ~2 alias contexts each), that's a `JSON.stringify(entire-cache) + synchronous localStorage.setItem` firing several times per second under polling. Coalesces to one save per 2s window — closer to one save per *full* poll cycle.

This **doesn't fix the nav-stuck bug** (CPU profile via CDP showed 83% idle during the stuck period — main thread isn't blocked) but reduces main-thread cost during normal polling.

### 2. ClusterAdmin uses `deduplicatedClusters`

`ClusterAdmin.tsx:13` destructured `clusters` (raw, 22 reachable for the user) instead of `deduplicatedClusters` (12 unique servers). My Clusters already used the deduped set; both pages now agree on the count.

## Issue #7865 status (nav-stuck) — NOT fixed in this PR

Tried and ruled out as causes (none consistently fixed it):

- `startTransition` wrap around cluster subscriber notify → caused stuck spinner regression (PR #7833)
- Disabling `checkHealthProgressively` → didn't fully unblock
- Stubbing `useClusters()` subscriber → didn't unblock
- Bypassing `DndContext` → didn't unblock
- Stripping all card renders → didn't unblock
- Stubbing entire Dashboard body → DID unblock (commits in 5s)
- `flushSync` around navigate → didn't unblock
- Manual `popstate` dispatch from CDP at +2.5s → DID unblock once (racy)
- Adding popstate dispatch to Sidebar onClick → didn't unblock

**Empirical findings**:

- Bug is racy (\"sometimes 1 click works, sometimes 2 needed\" per user)
- CPU profile via CDP: 83% IDLE during stuck period — not scheduler starvation
- Source-page-specific: ClusterAdmin → Deploy = 105ms; Dashboard or My Clusters → Deploy = often stuck
- `history.pushState` updates URL within 100ms of click; React Router's location subscription doesn't propagate the change to `<Routes>`/`<Outlet>` for some navigations

The bug needs deeper React Router internals investigation than I could resolve in this session.

## Test plan
- [ ] ClusterAdmin stat block shows same cluster count as My Clusters
- [ ] Cluster cards still update from polling (throttle is on the save, not the notify)
- [ ] Refresh spinner still toggles correctly
- [ ] No regression on other dashboards

### DIFF
--- a/web/src/components/cluster-admin/ClusterAdmin.tsx
+++ b/web/src/components/cluster-admin/ClusterAdmin.tsx
@@ -10,10 +10,14 @@ const STORAGE_KEY = 'kubestellar-cluster-admin-cards'
 const DEFAULT_CARDS = getDefaultCards('cluster-admin')
 
 export function ClusterAdmin() {
-  const { clusters: rawClusters, isLoading, isRefreshing, lastUpdated, refetch, error } = useClusters()
+  // Use deduplicatedClusters so the stat blocks count physical clusters, not
+  // kubeconfig contexts. Multiple contexts (long path + short alias + per-user
+  // SA variants) commonly point to the same server — the raw list double-counts.
+  // My Clusters already uses the deduplicated set; both pages must agree.
+  const { deduplicatedClusters, isLoading, isRefreshing, lastUpdated, refetch, error } = useClusters()
   const { getStatValue: getUniversalStatValue } = useUniversalStats()
 
-  const clusters = rawClusters || []
+  const clusters = deduplicatedClusters || []
 
   // Use the centralised health state machine so these counts always agree
   // with the main cluster grid, sidebar stats and filter tabs (#5928).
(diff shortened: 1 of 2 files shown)


clubanderson:
Closing per user request — neither change addresses the actual nav-stuck bug (per CPU profile in #7865, main thread is 83% idle during stuck period, so the localStorage save throttle isn't the blocker). The ClusterAdmin dedup count is a real but unrelated bug — can be reopened separately if desired. Branch deleted.

PR #7926, opened 2026-04-14
fix: cache invalid OpenAI/Gemini API key validation results

### LINKED ISSUE #7923: `validateOpenAIKey` returns `(false, error)` for a 401, which prevents caching the result in `ValidateAllKeys` — invalid OpenAI keys are re-validated on every startup
## User Request

**Type:** bug
**Console Request ID:** 974eef63-fedc-4c33-ae53-72b5e8d71890

## Description

**Description**

In `validateOpenAIKey`:
```go
if resp.StatusCode == http.StatusUnauthorized {
    return false, fmt.Errorf("invalid API key")
}
```
It returns a **non-nil error** for 401. In `ValidateAllKeys` (`server_operations.go:446–458`):
```go
valid, err := s.validateAPIKey(provider)
if err != nil {
    // Network or other error - don't cache, will try again later
    slog.Error("API key validation error (will retry)", ...)
} else {
    cm.SetKeyValidity(provider, valid)
}
```
Because `validateOpenAIKey` returns an error for 401, the validity is **never cached** for an invalid OpenAI key. Every startup re-fires the `/v1/models` request. By contrast, `validateClaudeKey` correctly returns `(false, nil)` for 401 — meaning Claude invalid keys are cached and not re-checked.

**What happened:**
An invalid OpenAI API key causes a live outbound HTTP request on every kc-agent startup instead of using the cached `(false)` result.

**What I expected:**
A 401 from any provider API is a definitive "invalid key" signal and should return `(false, nil)` to allow caching, not `(false, error)`.

**Steps to reproduce:**
1. Configure an invalid OpenAI key.
2. Restart kc-agent 5 times.
3. Observe 5 outbound requests to `api.openai.com/v1/models` in network logs (vs. 0 for an invalid Claude key after the first startup).

**File/Line:** `pkg/agent/server_operations.go:511–513` vs `server_operations.go:484–485`

**User impact:**
Unnecessary outbound network requests on startup. In rate-limited environments this could trigger OpenAI's rate limiter and cause false "API unavailable" states for other valid API consumers.

---

xyaz1313:
Fixes #7923

## Problem

 and  return  for HTTP 401/403 responses. In , any non-nil error skips caching:

This means invalid OpenAI/Gemini keys trigger a live API request on every startup, while invalid Claude keys (which correctly return ) are cached after the first check.

## Fix

Changed both  and  to return  for definitive auth failures (401/403), matching the existing  pattern.

### DIFF
--- a/pkg/agent/server_operations.go
+++ b/pkg/agent/server_operations.go
@@ -509,7 +509,7 @@ func validateOpenAIKey(ctx context.Context, apiKey string) (bool, error) {
 		return true, nil
 	}
 	if resp.StatusCode == http.StatusUnauthorized {
-		return false, fmt.Errorf("invalid API key")
+		return false, nil // Invalid key - no error so it gets cached by ValidateAllKeys
 	}
 	body, readErr := io.ReadAll(resp.Body)
 	if readErr != nil {
@@ -537,7 +537,7 @@ func validateGeminiKey(ctx context.Context, apiKey string) (bool, error) {
 		return true, nil
 	}
 	if resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden {
-		return false, fmt.Errorf("invalid API key")
+		return false, nil // Invalid key - no error so it gets cached by ValidateAllKeys
 	}
 	body, readErr := io.ReadAll(resp.Body)
 	if readErr != nil {


clubanderson:
Superseded by #8025 which landed the same fix. Thanks for the report and the initial patch!

PR #7983, opened 2026-04-14
fix(agent): restart-backend resolves real port + console binary (#7945)

### LINKED ISSUE #7945: bug: `handleRestartBackend` kills and restarts the backend Go process from within the kc-agent, but the kc-agent itself is started by `startup-oauth.sh` with a watchdog — the restart races against the
## User Request

**Type:** bug
**Console Request ID:** fe228a74-a216-4eed-932a-0086365c011c

## Description

**Description**

`handleRestartBackend` (`server_operations.go` region, `server.go:443`) calls `s.startBackendProcess()` which internally manages `s.backendCmd`. The system startup script (`startup-oauth.sh`) also runs a watchdog on port 8080 via `./watchdog`. When `handleRestartBackend` kills the backend, the watchdog detects the exit and schedules its own restart after the configured delay. Meanwhile, `startBackendProcess` also starts a new instance immediately. Both restarts racing leads to two backend instances attempting to bind port 8080 — the second one fails with `address already in use`.

**What happened:**
Clicking "Restart Backend" in the Settings panel results in one instance of the Go backend starting immediately from kc-agent and a second instance attempted by the watchdog 5 seconds later. The second instance exits immediately with port conflict. The UI briefly shows "Backend restarting..." then a 503.

**What I expected:**
The restart should be coordinated: either the watchdog handles all restarts (kc-agent signals it), or `handleRestartBackend` signals the watchdog to restart instead of directly managing the backend process.

**Steps to reproduce:**
1. Start the system with `./startup-oauth.sh`.
2. Trigger "Restart Backend" from Settings panel.
3. Observe two startup log lines: one from kc-agent's `startBackendProcess`, one from the watchdog restart (5s later).
4. The second one exits: `listen tcp :8080: bind: address already in use`.

**User impact:**
The Settings → Restart Backend feature is unreliable when the full `startup-oauth.sh` startup mode is used. Users see a transient 5-second gap where the backend is available, then a crash of the duplicate process, then stabilization — but the UI may already be showing a broken state.

---

clubanderson:
## The issue's analysis is wrong — here is what actually breaks

Issue #7945 speculates a "watchdog restart race" in `/restart-backend`. That is not what is happening. The watchdog (`cmd/console/watchdog.go:66-69`) is a **reverse proxy** that serves a "Reconnecting..." fallback page while the backend cycles; it never restarts anything. The real bugs are in `pkg/agent/server_http.go` — the kc-agent endpoint that the frontend POSTs to (`web/src/components/layout/Layout.tsx:131`) and that `selfUpdateFallback` (`pkg/agent/update_checker.go:778`) also calls.

Three real problems under the watchdog architecture:

1. **Wrong port killed.** `killBackendProcess` hard-coded `lsof -ti :8080 -sTCP:LISTEN`. In watchdog deployments port 8080 is the reverse-proxy listener; the backend runs on `BACKEND_PORT` (default 8081, set by `startup-oauth.sh:240,449`). The old code killed the watchdog and left the real backend alive — the opposite of the intent.

2. **Wrong binary re-execed.** `startBackendProcess` re-execed `os.Executable()`. In the kc-agent process that is the kc-agent binary, not `cmd/console`. The spawn then tried to bind kc-agent's own port (8585) and failed. The `KC_DEV_MODE=1` branch (`go run ./cmd/console`) was fine; the production branch silently restarted the wrong process.

3. **Wrong health target.** `checkBackendHealth` (and `waitForBackendHealth` in `update_checker.go`) hit a hard-coded `http://127.0.0.1:8080/health`, i.e. the watchdog, not the backend. Under the watchdog that returns "ok" even when the real backend is down, producing false-positive health.

## What changed

New helper `resolveBackendPort()` with three tiers:
1. `BACKEND_PORT` env var (set by `startup-oauth.sh`).
2. `backendPortWatchdogMode` (8081) if `/tmp/.kc-watchdog.pid` exists.
3. `backendPortLegacyDefault` (8080) for no-watchdog deployments.

All port literals are named constants in a dedicated block in `pkg/agent/server.go` (duplicated from `cmd/console/watchdog.go` because `pkg/agent` cannot import the main package — noted in the comment).

- `killBackendProcess`: `lsof -ti :<resolveBackendPort> -sTCP:LISTEN`, doc comment updated.
- `startBackendProcess`: new `resolveConsoleBinary()` search order — `KC_CONSOLE_BINARY` env var → sibling of `os.Executable()` (brew installs) → `exec.LookPath("console")` → error out rather than silently self-exec the wrong binary. `BACKEND_PORT` is injected into the child env via `envWithBackendPort()` which **dedupes** any inherited `BACKEND_PORT=…` entry.
- `checkBackendHealth` + `waitForBackendHealth`: now use `backendHealthURL()` which builds from the resolved port.

`selfUpdateFallback` in `update_checker.go` calls the same kill/restart callbacks through `UpdateCheckerConfig.RestartBackend` / `KillBackend`, so the brew-install auto-update path inherits the fix for free. Previously that path was silently leaving the old backend running after updates.

No frontend changes — `Layout.tsx:131` still POSTs to `/restart-backend` and that endpoint still returns `{success, killed, healthy}`.

## Tests added

- `TestResolveBackendPort` — subtests for legacy default, watchdog PID-file present, env-var override, and garbage env var falling through. Uses a package-level `watchdogPidFileStat` hook so the test never touches `/tmp`.
- `TestBackendHealthURL` — verifies the /health URL is built from the resolved port, not a stale constant.
- `TestEnvWithBackendPort` — guards that exactly one `BACKEND_PORT=…` entry ends up in the child env slice.

Existing `TestServer_HandleRestartBackend_*` tests still pass unchanged.

## Local verification

```
go build ./...         # pass
go vet ./...           # pass
go test ./pkg/agent/...# pass (72s)
```

## Closes

Fixes #7945

_AI-generated by Claude Code._

### DIFF
--- a/.github/workflows/update-guard.yml
+++ b/.github/workflows/update-guard.yml
@@ -83,8 +83,13 @@ jobs:
             echo "PASS: Atomic CAS for mutual exclusion present"
           fi
 
-          # 2. Defer cleanup of updating flag
-          if ! grep -q "defer atomic.StoreInt32" "$CHECKER"; then
+          # 2. Defer cleanup of updating flag.
+          # The cleanup was refactored from an inline `defer atomic.StoreInt32`
+          # into a local `cleanup := func() { atomic.StoreInt32(&uc.updating, 0); … }`
+          # closure that's deferred from every start path. Verify both:
+          # (a) cleanup closure exists and clears the updating flag, and
+          # (b) at least one `defer cleanup()` call site exists.
+          if ! grep -q "atomic.StoreInt32(&uc.updating, 0)" "$CHECKER" || ! grep -q "defer cleanup()" "$CHECKER"; then
             echo "FAIL: Missing defer cleanup of updating flag"
             ERRORS=$((ERRORS + 1))
           else
--- a/pkg/agent/server.go
+++ b/pkg/agent/server.go
@@ -31,10 +31,21 @@ const (
 	stabilizationDelay    = 3 * time.Second
 	startupDelay          = 500 * time.Millisecond
 	metricsHistoryTick    = 10 * time.Minute
-	agentFileMode         = 0600
-	defaultHealthCheckURL = "http://127.0.0.1:8080/health"
-	maxQueryLimit         = 1000    // Upper bound for client-supplied limit query parameter
-	maxRequestBodyBytes   = 1 << 20 // 1MB upper bound for request body reads
+	agentFileMode = 0600
+
+	// Backend port resolution constants (see resolveBackendPort in server_http.go).
+	// These are duplicated from cmd/console/watchdog.go because pkg/agent cannot
+	// import the main package. Keep them in sync with watchdog.go.
+	backendPortWatchdogMode  = 8081               // watchdog (8080) proxies -> backend (8081)
+	backendPortLegacyDefault = 8080               // no-watchdog deployments: backend binds 8080 directly
+	watchdogPidFilePath      = "/tmp/.kc-watchdog.pid"
+	backendHealthScheme      = "http"
+	backendHealthHost        = "127.0.0.1"
+	backendHealthPath        = "/health"
+	backendPortEnvVar        = "BACKEND_PORT"
+
+	maxQueryLimit       = 1000    // Upper bound for client-supplied limit query parameter
+	maxRequestBodyBytes = 1 << 20 // 1MB upper bound for request body reads
 
 	// missionExecutionTimeout is the maximum wall-clock time a single mission
 	// chat execution (AI provider call) is allowed to run before the context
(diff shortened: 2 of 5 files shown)


clubanderson:
Pushed a small CI follow-up: `update-guard.yml` was grepping for the literal `defer atomic.StoreInt32` but the cleanup was refactored into a `cleanup` closure (`defer cleanup()`). The functional invariant still holds — main has been failing this check silently because the workflow only fires on PRs that touch a handful of update-related files. Updated the grep to check for both `atomic.StoreInt32(&uc.updating, 0)` in the body AND at least one `defer cleanup()` call site.

Issue #6452, opened 2026-04-11
E2E Tests Hardcode Port 8080 While Playwright Config Uses Port 5174


arnavgogia20:
## User Request

**Type:** bug
**Target:** Console Application
**Submitted by:** @arnavgogia20
**Console Request ID:** 975746cd-36f5-48bd-9cf6-d7fbbf424341

## Area

Testing / DX

---

## Problem

The Playwright configuration in `playwright.config.ts` defines:

```typescript
baseURL: process.env.PLAYWRIGHT_BASE_URL || 'http://localhost:5174',

webServer: {
  command: 'npm run dev -- --port 5174',
  url: 'http://localhost:5174',
}
```

However, the **mission-control E2E tests** hardcode URLs pointing to **port 8080**:

```typescript
// mission-control-e2e.spec.ts
await page.goto('http://localhost:8080/login')
await page.goto('http://localhost:8080')
await page.goto('http://localhost:8080/?mission-control=open')
await page.goto('http://localhost:8080/?browse=missions')

// mission-control-dry-run.spec.ts
await page.goto('http://localhost:8080/login')
await page.goto('http://localhost:8080')

// mission-control-stress.spec.ts
await page.goto('http://localhost:8080/login')
await page.goto('http://localhost:8080')
```

---

## Consequences

- Running:

```bash
npx playwright test
```

spins up a dev server on **port 5174**, but tests attempt to navigate to **port 8080**

- Result: **Connection refused errors**

- Tests only pass when:
  - Running `startup-oauth.sh`
  - Backend is manually available on `localhost:8080`

This creates an inconsistent and fragile testing setup.

---

## Reproduction Steps

1. Run:

```bash
npx playwright test e2e/mission-control-e2e.spec.ts
```

2. Observe:
   - Playwright starts server at `http://localhost:5174`
   - Tests navigate to `http://localhost:8080`
   - Tests fail with **connection refused**

---

## Expected Behavior

All tests should:

- Use **relative URLs**
- Rely on Playwright’s `baseURL` configuration
- Work seamlessly across environments:
  - Local dev server (`5174`)
  - OAuth server (`8080`)
  - CI environments

---

## Suggested Fix

Replace all hardcoded URLs with **relative paths**.

### Example Refactor

```diff
- await page.goto('http://localhost:8080/login')
+ await page.goto('/login')

- await page.goto('http://localhost:8080')
+ await page.goto('/')

- await page.goto('http://localhost:8080/?mission-control=open')
[...]

PR #8055, opened 2026-04-15
✨ Add Playwright UX scanning 

xonas1101:
### 📝 Summary of Changes

Added a script that runs UX related playwright tests and persists the results on errors into a test-result folder (not tracked). 
The md file is a prompt for AI agent to run against the test-result folder and open issues accordingly.

---

### Changes Made

- [x] Updated AI agent issue generation through UX behavior
- [ ] Refactored ...
- [ ] Fixed ...
- [ ] Added tests for ...

---

### Checklist

Please ensure the following before submitting your PR:

- [x] I used a coding agent (Claude Code, Copilot, Gemini, or Codex) to generate/review this code
- [x] I have reviewed the project's contribution guidelines
- [ ] New cards target [console-marketplace](https://github.com/kubestellar/console-marketplace), not this repo
- [ ] isDemoData is wired correctly (cards show Demo badge when using demo data)
- [ ] I have written unit tests for the changes (if applicable)
- [x] I have tested the changes locally and ensured they work as expected
- [x] All commits are signed with DCO (`git commit -s`)

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,7 +44,9 @@ coverage/
 web/playwright-report/
 web/test-results/
 web/blob-report/
+web/playwright/.cache/
 web/playwright/.auth/
+web/auth.json
 
 # Temp directories
 a2a-rebase/
@@ -70,6 +72,7 @@ vendor/
 *.bak
 web/.netlify/
 web/e2e/test-results/
+web/e2e/user-flows/test-results/
 
 # Storybook
 web/storybook-static/
--- a/docs/qa/AI-UX-ISSUE-AGENT-BRIEF.md
+++ b/docs/qa/AI-UX-ISSUE-AGENT-BRIEF.md
@@ -1,0 +1,83 @@
+# AI UX Issue Agent Brief
+
+Use this file as the operating brief for an AI agent that finds and reports UX issues from Playwright results.
+
+## Goal
+
+Identify and report only browser-reproduced UX issues from the UX scan output.
+
+Do not report code-scan-only findings.
+
+## Inputs
+
+1. Run the UX scan first:
+
+```bash
+./scripts/run-ux-scan.sh
+```
+
+Notes:
+- For localhost URLs, the script auto-starts a preview server when needed.
+- To disable auto-start behavior, set `UX_SCAN_AUTOSTART_SERVER=false`.
+
+2. Read these files:
+- `web/test-results/ux-scan/ux-findings.json`
+- `web/test-results/ux-scan/ux-summary.md`
+- `web/test-results/ux-scan/playwright-report/index.html`
+
+## Hard Rules
+
+1. Only use findings from `ux-findings.json`.
+2. Only report findings where `source.reproducedInBrowser` is `true`.
+3. If no findings exist, report "No browser-reproduced UX issues found" and stop.
+4. Every reported issue must include route or spec source, reproduction steps, expected, actual, and severity.
+5. Do not invent failures that are not present in Playwright artifacts.
+
+## Issue JSON Contract
+
+Use this shape for each issue:
+
+```json
+{
+  "issue": "Short title",
+  "steps": ["step 1", "step 2"],
+  "expected": "What should happen",
+  "actual": "What happened",
+  "severity": "high | medium | low"
+}
+```
+
+## Agent Procedure
+
+1. Parse `web/test-results/ux-scan/ux-findings.json`.
+2. Filter to valid browser-reproduced findings.
+3. Normalize severity:
+- `timedOut` or console/exception failures => `high`
+- assertion/navigation/visibility failures => `medium`
+- cosmetic-only regressions => `low`
+4. Deduplicate by issue title + source file + line.
+5. Append results to the "Reported Issues" section in this file.
+
+
+## Reported Issues
+
+Append new findings below this line.
+
+---
+
+### Report Batch: YYYY-MM-DD HH:MM UTC
+
+- Status: Pending
+- Total findings: 0
+
+#### Issue 1
+
+```json
+{
+  "issue": "",
+  "steps": [],
+  "expected": "",
+  "actual": "",
+  "severity": ""
+}
+```
--- a/docs/qa/AI-UX-ISSUE-AGENT-BRIEF.md
+++ b/docs/qa/AI-UX-ISSUE-AGENT-BRIEF.md
@@ -56,12 +56,15 @@ Use this shape for each issue:
 - assertion/navigation/visibility failures => `medium`
 - cosmetic-only regressions => `low`
 4. Deduplicate by issue title + source file + line.
-5. Append results to the "Reported Issues" section in this file.
+5. Write results to `web/test-results/ux-scan/reported-issues.md`.
+
+Do not modify this brief file. It is checked into git and should remain stable.
 
 
 ## Reported Issues
 
-Append new findings below this line.
+This section is intentionally left as an example template.
+Write real run output to `web/test-results/ux-scan/reported-issues.md`.
 
 ---
 
--- a/scripts/run-ux-scan.sh
+++ b/scripts/run-ux-scan.sh
@@ -11,10 +11,14 @@ FINDINGS_JSON="${UX_DIR}/ux-findings.json"
 SUMMARY_MD="${UX_DIR}/ux-summary.md"
 LEGACY_RAW_JSON="${WEB_DIR}/e2e/user-flows/test-results/ux-scan/ux-raw-results.json"
 
-BASE_URL="${PLAYWRIGHT_BASE_URL:-http://localhost:5174}"
+DEFAULT_BASE_URL="http://localhost:8080"
+WAIT_ON_TIMEOUT_MS=120000
+
+BASE_URL="${PLAYWRIGHT_BASE_URL:-${DEFAULT_BASE_URL}}"
 STORAGE_STATE="${PLAYWRIGHT_STORAGE_STATE:-auth.json}"
 AUTO_START_SERVER="${UX_SCAN_AUTOSTART_SERVER:-true}"
 PREVIEW_PID=""
+PREVIEW_LOG=""
 
 mkdir -p "${UX_DIR}"
 
@@ -24,7 +28,9 @@ cleanup() {
   if [[ -n "${PREVIEW_PID}" ]]; then
     kill "${PREVIEW_PID}" >/dev/null 2>&1 || true
   fi
-  rm -f /tmp/ux-scan-preview.log
+  if [[ -n "${PREVIEW_LOG}" ]]; then
+    rm -f "${PREVIEW_LOG}" || true
+  fi
 }
 
 trap cleanup EXIT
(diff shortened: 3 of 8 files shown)


clubanderson:
Hi @xonas1101 — thanks for putting this together! UX scanning is something we care a lot about, and the AI-agent brief in particular is a nice touch. I went through all 8 files carefully. Posting a review here before this can move forward.

## Summary

- **Security:** Clean. No command injection, no remote calls, no credential exfiltration, no prompt injection in the agent brief. DCO is signed.
- **Code quality / fit:** Several blocking concerns, the biggest being that `web/playwright.config.ts` looks like it was regenerated from `npx create-playwright` boilerplate rather than edited in place — that wipes out a lot of intentional config the project depends on.
- **Scope:** This overlaps significantly with `nightly-ux-journeys.yml`, which already does Playwright UX sweeps in demo mode. We should reconcile the two before merging anything new.

## Blocking — `web/playwright.config.ts` regression (–134 / +78)

This file is a near-complete rewrite, and most of the deletions are load-bearing config the project added deliberately. The diff shows your version added the boilerplate `// import dotenv ...` and `/* Run your local dev server before starting the tests */ // webServer: ...` placeholder comments that ship with `npx create-playwright`, which strongly suggests the file was regenerated rather than patched. Specific regressions:

1. **Removes the `webServer` block that auto-starts the Go backend on `:8080`.** This was added in #6452/#6474 with a long explanatory comment (which you also deleted) explaining that in production the Go backend serves both API and frontend on `:8080`, and Playwright must match the real deployment topology. Removing it means `npm run test:e2e` no longer works locally without manually starting a server.
2. **Changes `baseURL` from `http://localhost:8080` back to commented-out / implicit.** Same reason as above — this breaks the `startup-oauth.sh` workflow that the project standardized on.
3. **Removes `mobile-chrome` and `mobile-safari` projects entirely.** The repo intentionally tests against mobile viewports.
4. **Removes the CI reporter array** (`blob`, `json`, `junit`, `github`) and the custom `./e2e/helpers/ux-reporter.ts` reporter — falls back to bare `'html'`. This breaks CI artifact collection and the existing UX reporter pipeline.
5. **Hardcodes `workers: 1`** instead of `process.env.CI ? 4 : '50%'`. That's a 4× slowdown in CI and ~50× slowdown locally on multi-core boxes.
6. **Changes `retries` from `CI ? 1 : 0` to `CI ? 2 : 0`** with no rationale — doubles CI time on flakes.
7. **Removes shared `screenshot: 'only-on-failure'`, `video: 'retain-on-failure'`, and `viewport: { width: 1280, height: 720 }`** from the `use` block.
8. **Adds an explicit `testMatch` allowlist** that hard-codes spec filenames (`Login.spec.ts`, `Dashboard.spec.ts`, etc.). Any new spec a contributor adds in the future will silently not run unless they remember to update the allowlist. The previous config used `testIgnore` (allow by default), which is the right default for a growing test suite.
9. **Inconsistent style:** uses semicolons throughout (`defineConfig({...});`), but the rest of the codebase and your own new file `ux-scan.config.ts` are no-semi. Project lint/prettier will probably strip these on commit.
10. **Defensive cast `(globalThis as { process?: { env?: ... } }).process?.env?.CI`** — this is a Node config file, not browser code. `process.env.CI` works fine. The cast adds noise.

**Ask:** Please revert `web/playwright.config.ts` and instead either (a) leave it alone entirely and only add the new `ux-scan.config.ts`, or (b) make a minimal edit to add the new spec files to existing test discovery. The new UX sweep already has its own dedicated config (`e2e/user-flows/ux-scan.config.ts`), so it doesn't need to touch the main config at all.

## Blocking — Scope / fit

The repo already has a nightly UX sweep: `.github/workflows/nightly-ux-journeys.yml`. It runs Playwright UX journey tests against demo mode, generates findings, and files issues — same idea as this PR. Before adding parallel infrastructure, can you walk through the existing workflow and tell us:

- What gap does your script fill that `nightly-ux-journeys.yml` doesn't?
- Should the new `post-login-dashboard-ux.spec.ts` be added to that nightly job instead of running via a new local-only shell script?
- Should `AI-UX-ISSUE-AGENT-BRIEF.md` reference the existing nightly artifacts (which already exist in CI) instead of requiring contributors to run a local shell script first?

I think the spec file and the agent brief are valuable on their own — it's just the run-script + config-rewrite layer that duplicates existing infra.

## Code quality nits (non-blocking, address with the rewrite)

1. **`scripts/run-ux-scan.sh`** — Hardcoded `/tmp/ux-scan-preview.log` is a predictable path; use `mktemp` for symlink safety. Hardcoded `120000` (ms) timeout on `npx wait-on` should be a named constant per [CLAUDE.md](../blob/main/CLAUDE.md#no-magic-numbers). The `node -e \"...\"` URL parsing could just be a few lines of bash with `${BASE_URL#*://}` parameter expansion — fewer moving parts and no Node dependency on the shebang path.
2. **`web/e2e/user-flows/ux-scan.config.ts`** — Magic numbers `240_000`, `20_000`. CLAUDE.md is strict about this: every numeric literal needs a named constant with a comment. Your spec file `post-login-dashboard-ux.spec.ts` actually does this correctly (`DASHBOARD_LOAD_TIMEOUT_MS` etc.) — please apply the same pattern to the config.
3. **`AI-UX-ISSUE-AGENT-BRIEF.md`** — The \"Append results to the 'Reported Issues' section in this file\" instruction tells the agent to mutate a checked-in file. That's a recipe for endless dirty PRs. Better to write findings to `web/test-results/ux-scan/reported-issues.md` (which `.gitignore` already covers) and link to it from the brief.
4. **`post-login-dashboard-ux.spec.ts`** — The login flow sets `localStorage.setItem('token', 'demo-token')` directly. There's an existing helper for this in the e2e suite (search for `setupDemoMode` or similar in `web/e2e/helpers/`); please reuse it so demo-mode setup stays consistent across specs.
5. **`web/package.json` script** — The new `test:e2e:ux-sweep` hardcodes `PLAYWRIGHT_BASE_URL=http://localhost:5174`, but the rest of the project standardized on `:8080` (Go backend serves both). If we keep this script, it should match.
6. The relative `outputDir: '../../test-results/ux-scan/artifacts'` in `ux-scan.config.ts` is fragile — if the file ever moves, the path silently breaks. Consider deriving it from `__dirname` or putting it under a top-level paths constant.

## Security review (no findings)

I checked all 8 files for command injection, credential leaks, prompt-injection in the agent brief, and remote calls. All clean:
- `scripts/run-ux-scan.sh` uses `set -euo pipefail`, all variables are quoted, no `eval` or `curl|bash`, the `cleanup` trap only kills the child it spawned.
- `AI-UX-ISSUE-AGENT-BRIEF.md` is plain ASCII, no hidden Unicode, no \"ignore safety rules\" instructions, and explicitly tells the agent to only report browser-reproduced findings and not invent failures. This is a well-written prompt.
- `post-login-dashboard-ux.spec.ts` uses a fake `demo-token`, never logs captured data, never hits production endpoints.
- `.gitignore` additions are all legitimate (cache dirs, the `auth.json` storage state file, additional test-results paths).

So from a security standpoint there is nothing here to fix.

---

**Do not merge until the playwright.config.ts regression is reverted and the scope question vs. `nightly-ux-journeys.yml` is resolved.** Holding for @clubanderson on the scope/fit decision.

Thanks again for the contribution — the spec file and the agent brief are both genuinely useful, and once we untangle them from the config rewrite I think a slimmed-down version of this PR can land quickly.

clubanderson (changes requested):
Thanks @xonas1101 for the follow-up commit — the UX-scan side of the feedback is largely addressed:

- `test:e2e:ux-sweep` now targets `http://localhost:8080` to match the Go-served topology
- `run-ux-scan.sh` uses `mktemp` for the preview log and extracts `WAIT_ON_TIMEOUT_MS`
- `ux-scan.config.ts` pulls out `UX_SWEEP_TIMEOUT_MS`, `EXPECT_TIMEOUT_MS`, and output paths into named constants
- `post-login-dashboard-ux.spec.ts` now uses the existing `setMode(page, 'demo')` helper instead of hand-rolled `localStorage.setItem` calls
- Agent brief no longer self-mutates — writes run output to `web/test-results/ux-scan/reported-issues.md`

Nice work on those.

## Still blocking — `web/playwright.config.ts` is unchanged and still regressed

The `99611f1b` follow-up commit touched 5 files but **did not touch `web/playwright.config.ts`**, so all three of the Copilot review findings on that file are still live. This file is the default config that ALL Playwright workflows fall back to, so the breakage is repo-wide, not scoped to the new UX scan.

### 1. `reporter: 'html'` breaks `playwright.yml` shard/merge pipeline (line 52)

The existing `.github/workflows/playwright.yml` uploads `web/blob-report/` as an artifact (line 105-106) and later runs `npx playwright merge-reports --reporter=html,json ./all-blob-reports` (line 153). With `reporter: 'html'` there is no blob reporter output, so every sharded Playwright run in CI will upload empty artifacts and `merge-reports` will have nothing to merge.

Restore a CI-aware reporter setup, e.g.:

```ts
reporter: isCI
  ? [
      ['blob', { outputDir: 'blob-report' }],
      ['html', { outputFolder: 'playwright-report' }],
      ['json', { outputFile: 'test-results/results.json' }],
      ['junit', { outputFile: 'test-results/junit.xml' }],
      ['github'],
    ]
  : [['html', { open: 'never' }], ['./e2e/helpers/ux-reporter.ts']],
```

### 2. `testIgnore: ['**/fullstack-smoke.spec.ts']` breaks `fullstack-e2e.yml` (line 41)

`.github/workflows/fullstack-e2e.yml` line 106 runs:

```bash
npx playwright test e2e/fullstack-smoke.spec.ts --project=chromium
```

That command uses the default `web/playwright.config.ts`. With this spec listed in `testIgnore`, playwright will refuse to run it and the fullstack smoke job fails. Remove `'**/fullstack-smoke.spec.ts'` from `testIgnore` — this workflow is the only place that spec runs, and it is explicitly exercising the Go-backend-serves-built-frontend production path, which is exactly what you're now aligning the UX scan with.

### 3. `baseURL` commented out (line 56)

```ts
use: {
  /* Base URL to use in actions like `await page.goto('')`. */
  // baseURL: 'http://localhost:3000',
```

The existing specs under `e2e/` (Login.spec.ts, Dashboard.spec.ts, Sidebar.spec.ts, smoke.spec.ts, etc.) rely heavily on relative navigation like `page.goto('/login')`. Without a `baseURL` those calls resolve against `about:blank` and fail immediately. Restore:

```ts
baseURL: process.env.PLAYWRIGHT_BASE_URL || 'http://localhost:8080',
```

(This also matches the port you correctly picked for the UX scan — so the default config and the UX-scan config stay consistent.)

### 4. `webServer` block deletion

The original config had a `webServer` entry that, when `PLAYWRIGHT_BASE_URL` is unset, launches `cd .. && go run .` on port 8080 to serve the real production topology (carrying explicit `#6452/#6474` fix markers). The new file replaces it with commented-out boilerplate:

```ts
/* Run your local dev server before starting the tests */
// webServer: {
//   command: 'npm run start',
//   url: 'http://localhost:3000',
//   reuseExistingServer: !process.env.CI,
// },
```

Local `npm run test:e2e` runs will hang on connect like they did pre-#6452/#6474. Restore the Go-backend `webServer` block.

### 5. Other losses in the rewrite

- `mobile-chrome` / `mobile-safari` projects removed — used by mobile smoke specs
- `timeout: 60000` → top-level timeout removed
- `expect.timeout: 10000` → removed
- `forbidOnly`, `retries`, `workers` swapped from `process.env.CI` to a new `isCI` local (cosmetic, fine, keep it)
- Lint error count went from 343 (main) to 361 (PR) — 18 new lint errors. Please run `npm run lint` locally and resolve.

## Suggested fix

The cleanest path forward is to **revert `web/playwright.config.ts` to main** and only add the UX-scan bits via the dedicated `web/e2e/user-flows/ux-scan.config.ts` (which is already opt-in via `npm run test:e2e:ux-sweep` and the standalone shell script). The default `playwright.config.ts` should NOT change at all for this PR — the UX sweep is already wired through its own config file and script.

If you do want to add `user-flows/**` to the default config's `testMatch`, please layer that surgically on top of the current file rather than replacing it.

## Re-review checklist

- [ ] `web/playwright.config.ts` restored to main's content (or surgically updated without losing blob reporter, baseURL, webServer, mobile projects, timeouts, or fullstack-smoke inclusion)
- [ ] `.github/workflows/playwright.yml` (sharded runs, merge-reports) still produces blob artifacts
- [ ] `.github/workflows/fullstack-e2e.yml` (`fullstack-smoke.spec.ts --project=chromium`) still runs
- [ ] Lint delta is zero (`npm run lint` count matches main)
- [ ] Scope reconciled with `nightly-ux-journeys.yml` (or documented why both are needed)

Everything else you did on this round looks good — appreciate the careful rework on the UX-scan pieces. Just need the default config reverted and we're close.

clubanderson:
Hi @xonas1101 — closing this in favor of #9225, which is a clean-room version that incorporates all the feedback from both review rounds.

The only delta vs your final commit (`963e5bd`) is in `web/playwright.config.ts`: your branch had inadvertently reverted `testIgnore` from `[]` back to the old `['**/auth.setup.ts', ...]` array that was removed in #9076 as dead code. Everything else — the UX scan spec, config, shell script, agent brief, gitignore additions, package.json script, and PLAYWRIGHT.md docs — is taken directly from your work.

Credit is noted in the PR description. Thank you for the contribution!

PR #8083, opened 2026-04-15
fix(agent): drain Statuspage response body after JSON decode

### LINKED ISSUE #8069: bug: `checkStatuspageHealth` reads the entire response body into a JSON decoder but never drains it, potentially holding a persistent HTTP connection open
## User Request

**Type:** bug
**Console Request ID:** e91b3913-ad9f-44ac-8091-f9d6ef8500e0

## Description

**Description:**
`checkStatuspageHealth` in `pkg/agent/server_operations.go` (line 648) uses `json.NewDecoder(resp.Body).Decode(&data)`. If the JSON decoder stops reading early (e.g., after finding the `status` field), the rest of the response body is not consumed, preventing the underlying TCP connection from being returned to the pool.

**What happened:**
When the Statuspage payload is large, the connection is held open until Go's `http.Client` timeout fires, slowly exhausting the connection pool under high request rates.

**What I expect:**
After decoding, the remaining body should be explicitly drained: `io.Copy(io.Discard, resp.Body)`.

**Steps to reproduce:**
1. Mock a Statuspage endpoint that returns a large response body after the `status` field.
2. Call `/providers/health` at a high rate.
3. Monitor open TCP connections — observe connections building up and not closing.

**User impact:**
Under sustained use (e.g., frequent health polling), the agent slowly exhausts TCP connections and begins failing provider health checks.

---

mvanhorn:
## Summary

Addresses #8069. `checkStatuspageHealth` in `pkg/agent/server_operations.go` was calling `json.NewDecoder(resp.Body).Decode(&data)`, which stops reading as soon as the top-level `status` field is consumed. Statuspage payloads include `components`, `incidents`, `scheduled_maintenances`, and `page` fields after `status`, so those bytes stayed in the socket buffer and the underlying TCP connection could not be returned to the keep-alive pool until the HTTP client timeout fired.

Under sustained `/providers/health` polling this slowly exhausted the connection pool.

## Fix

One line: `io.Copy(io.Discard, resp.Body)` after `Decode`, before `defer resp.Body.Close()` runs. The error is intentionally discarded (`_, _ =`) because we're already past the meaningful JSON and only care about freeing the connection.

## Test plan

- [x] `go test ./pkg/agent/ -run TestCheckStatuspageHealth` — all pre-existing cases still pass
- [x] Added `TestCheckStatuspageHealth_DrainsBody` — uses a realistic multi-field Statuspage response, asserts a second call on the same client completes in <2s as a proxy for connection reuse, and sanity-checks the server actually wrote bytes
- [x] `go build ./...` clean
- [x] `go vet ./pkg/agent/...` clean

Closes #8069

### DIFF
--- a/pkg/agent/server_operations.go
+++ b/pkg/agent/server_operations.go
@@ -708,6 +708,11 @@ func checkStatuspageHealth(client *http.Client, apiURL string) string {
 	if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
 		return "unknown"
 	}
+	// Drain any unread portion of the body so the underlying TCP connection
+	// can be returned to the pool instead of held open until the HTTP client
+	// timeout fires. Statuspage payloads include fields we don't decode,
+	// which left the decoder stopped mid-stream.
+	_, _ = io.Copy(io.Discard, resp.Body)
 
 	switch data.Status.Indicator {
 	case "none":
--- a/pkg/agent/server_test.go
+++ b/pkg/agent/server_test.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


clubanderson:
Hi @mvanhorn — thank you so much for the clean fix and the test! Really appreciate the effort, especially the `TestCheckStatuspageHealth_DrainsBody` test that proves connection reuse.

Unfortunately this is a duplicate: PR #8082 (created ~1 minute before this one) bundles a fix for #8069 along with five other bugs from the same batch report, using the same approach (`io.Copy(io.Discard, resp.Body)` added to the deferred cleanup). Because #8082 ships with admin merge today, this PR will end up with a merge conflict against the exact same hunk.

I'm going to close this as a duplicate to keep the tree tidy — but please don't be discouraged. The fix was correct and your test case is actually *better* than what we shipped in #8082 (we didn't write an explicit drain test). If you'd like, you could open a follow-up PR adding just the test (`TestCheckStatuspageHealth_DrainsBody`) once #8082 merges — that would be a genuinely useful contribution and I'd merge it immediately.

Thanks again for jumping in — please keep an eye on the issue tracker, there's plenty more to fix.

Issue #2654, opened 2026-03-15
bug: SSE streams do not reconnect when connection drops


aaradhychinche-alt:
## User Request

**Type:** bug
**Target:** Console Application
**Submitted by:** @aaradhychinche-alt
**Console Request ID:** e34843c0-45a9-45c9-aa5f-ff7a08c3e926

## Description

What happened:
Several hooks in the console rely on Server-Sent Events (SSE) streams to receive updates from the backend (for example pod status, deployments, jobs, and issues). When the SSE connection drops due to a backend restart, network interruption, or cluster reconnect, the stream closes and the UI stops receiving updates. The current implementation does not attempt to reconnect the SSE stream, which causes the dashboard to continue displaying stale data.

What I expected:
If an SSE connection drops unexpectedly, the client should attempt to reconnect automatically so that the UI continues receiving updates and reflects the latest cluster state.

Steps to reproduce:
1. Start the console and connect it to a cluster.
2. Open a dashboard that displays workload-related cards.
3. Restart the backend service or temporarily interrupt the network connection.
4. Observe that the SSE stream closes and the UI no longer receives updates, even after the connection becomes available again.

Relevant hooks:
web/src/hooks/mcp/workloads.ts
web/src/hooks/mcp/usePods.ts
web/src/hooks/mcp/usePodIssues.ts
web/src/hooks/mcp/useDeploymentIssues.ts
web/src/hooks/mcp/useDeployments.ts

Suggested improvement:
Implement automatic reconnection logic for SSE streams with exponential backoff so that the client reconnects when the stream closes unexpectedly. This would prevent stale data and improve reliability of real-time dashboard updates.

---
*This issue was automatically created from the KubeStellar Console.*

PR #8142, opened 2026-04-15
🐛 fix: thundering herd websocket reconnect exponential backoff

khushal-winner:
- Add exponential backoff (2s → 4s → 8s → 16s → 30s max) with 0-1000ms jitter
- Implement maximum retry limits (5 attempts) with graceful failure
- Update useClusterProgress, useUpdateProgress, useAIPredictions, and useActiveUsers hooks
- Add proper unmounted guards to prevent timer leaks
- Set failed state when max reconnect attempts exceeded
- Follow existing useExecSession.ts pattern for consistency

---

### Changes Made

- [ ] Updated ...
- [ ] Refactored ...
- [x] Fixed ...
- [ ] Added tests for ...

---

### Checklist

Please ensure the following before submitting your PR:

- [x] I used a coding agent (Claude Code, Copilot, Gemini, or Codex) to generate/review this code
- [x] I have reviewed the project's contribution guidelines
- [ ] New cards target [console-marketplace](https://github.com/kubestellar/console-marketplace), not this repo
- [ ] isDemoData is wired correctly (cards show Demo badge when using demo data)
- [ ] I have written unit tests for the changes (if applicable)
- [x] I have tested the changes locally and ensured they work as expected
- [x] All commits are signed with DCO (`git commit -s`)

---

### Screenshots or Logs (if applicable)

---

### 👀 Reviewer Notes

_Add any special notes for the reviewer here_

### DIFF
--- a/web/src/hooks/useAIPredictions.ts
+++ b/web/src/hooks/useAIPredictions.ts
@@ -2,7 +2,8 @@ import { useState, useEffect } from 'react'
 import type {
   AIPrediction,
   AIPredictionsResponse,
-  PredictedRisk } from '../types/predictions'
+  PredictedRisk
+} from '../types/predictions'
 import { getPredictionSettings, getSettingsForBackend } from './usePredictionSettings'
 import { getDemoMode } from './useDemoMode'
 import { isAgentUnavailable, reportAgentDataSuccess, reportAgentDataError } from './useLocalAgent'
@@ -10,7 +11,13 @@ import { setActiveTokenCategory, clearActiveTokenCategory } from './useTokenUsag
 import { fullFetchClusters, clusterCache } from './mcp/shared'
 
 import { LOCAL_AGENT_WS_URL, LOCAL_AGENT_HTTP_URL } from '../lib/constants'
-import { FETCH_DEFAULT_TIMEOUT_MS, AI_PREDICTION_TIMEOUT_MS, WS_RECONNECT_DELAY_MS, UI_FEEDBACK_TIMEOUT_MS, RETRY_DELAY_MS } from '../lib/constants/network'
+import { FETCH_DEFAULT_TIMEOUT_MS, AI_PREDICTION_TIMEOUT_MS, UI_FEEDBACK_TIMEOUT_MS, RETRY_DELAY_MS } from '../lib/constants/network'
+
+// WebSocket reconnection with exponential backoff
+const WS_RECONNECT_BASE_DELAY_MS = 2_000  // Base delay for reconnection attempts
+const WS_RECONNECT_MAX_DELAY_MS = 30_000   // Maximum delay between reconnection attempts
+const MAX_WS_RECONNECT_ATTEMPTS = 5        // Maximum reconnection attempts before giving up
+const BACKOFF_JITTER_MAX_MS = 1_000        // Random jitter to avoid thundering herd
 
 const AGENT_HTTP_URL = LOCAL_AGENT_HTTP_URL
 const POLL_INTERVAL_MS = 30_000 // Poll every 30 seconds as fallback
@@ -28,7 +35,8 @@ const DEMO_AI_PREDICTIONS: AIPrediction[] = [
     confidence: 78,
     generatedAt: new Date().toISOString(),
     provider: 'claude',
-    trend: 'worsening' },
+    trend: 'worsening'
+  },
   {
     id: 'demo-ai-2',
     category: 'anomaly',
@@ -39,7 +47,8 @@ const DEMO_AI_PREDICTIONS: AIPrediction[] = [
     reasonDetailed: 'Pod has restarted 4 times in the past 3 hours, with each restart occurring during traffic peaks. This suggests memory or CPU limits may be too low for peak load. Recommend increasing resource limits or implementing HPA.',
     confidence: 85,
     generatedAt: new Date().toISOString(),
-    provider: 'claude' },
+    provider: 'claude'
+  },
 ]
 
 // Singleton state - shared across all hook instances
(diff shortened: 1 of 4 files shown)


clubanderson:
Thanks for the PR, @khushal-winner. A few notes for the maintainer and a concern about the issue linkage:

## The `Fixes #2654` claim doesn't match the change

Issue #2654 is about **SSE stream reconnect** in these hooks:
```
web/src/hooks/mcp/workloads.ts
web/src/hooks/mcp/usePods.ts
web/src/hooks/mcp/usePodIssues.ts
web/src/hooks/mcp/useDeploymentIssues.ts
web/src/hooks/mcp/useDeployments.ts
```

It was **already fixed** on main by PR #2720 (commit `86923c77` — "Add auto-reconnect with exponential backoff to SSE streams"). The issue is closed.

This PR touches a completely different set of files (**WebSocket** hooks, not SSE):
- `web/src/hooks/useAIPredictions.ts`
- `web/src/hooks/useActiveUsers.ts`
- `web/src/hooks/useClusterProgress.ts`
- `web/src/hooks/useUpdateProgress.ts`

So merging this would not actually "fix #2654" — the auto-close keyword is misleading.

## The code change itself looks reasonable

The exponential-backoff-with-jitter pattern mirrors `useExecSession.ts`, constants are named (`WS_RECONNECT_BASE_DELAY_MS`, `WS_RECONNECT_MAX_DELAY_MS`, `MAX_WS_RECONNECT_ATTEMPTS`, `BACKOFF_JITTER_MAX_MS`), and the reset-on-open + attempt cap look correct. No security concerns in the diff (no new fetch URLs, eval, innerHTML, or auth-path changes).

## Ask before proceeding

1. Is there an actual reported thundering-herd incident on these WS hooks, or is this preventive hardening? If preventive, it should be filed as its own issue first so reviewers know what problem it solves.
2. If this is worth landing, please remove `Fixes #2654` from the description — that issue is fixed and unrelated.

Leaving this open for a human maintainer to decide per our external-contributor policy.

clubanderson (changes requested):
Thanks for tackling thundering herd — jitter, max cap, and reset-on-success are all correctly implemented, and the named constants are great. However, I need to request changes because the 5-attempt giveup introduces regressions in hooks whose whole purpose is to survive extended disconnects.

## Blocking issues

### 1. `useUpdateProgress` regression (critical)

This hook exists specifically to track backend restarts during self-upgrade, where the WebSocket is *expected* to drop for well over a minute (docker pull + build + restart). The old code had an explicit comment:

```ts
// Reconnect after 5 seconds (faster during restarts)
reconnectTimer = setTimeout(connect, WS_RECONNECT_MS)
```

With this PR, after `2 + 4 + 8 + 16 + 30 ≈ 60s` the hook silently gives up and logs `[UpdateProgress] Max reconnect attempts exceeded, giving up`. That's exactly the scenario this hook was built to handle, so the "Update Progress" card will go dark during every upgrade that takes longer than ~1 minute — which is most of them. Please keep `useUpdateProgress` retrying forever (capped at `WS_RECONNECT_MAX_DELAY_MS`), or at minimum raise the attempt cap to something like 100.

### 2. Same regression in the other three hooks

`useAIPredictions`, `useActiveUsers`, and `useClusterProgress` are long-lived background hooks running for the entire user session. Any backend maintenance window > ~60s (pod restart, rollout, node drain) will permanently break them until the user reloads the page — no UI indication, no recovery. The `useExecSession.ts` pattern you're following is a poor fit here: that hook is user-initiated with an explicit "Connect manually" affordance when it gives up. These four hooks have no such affordance.

**Recommendation:** thundering herd is fully mitigated by jitter + capped exponential backoff alone. The "give up after N attempts" piece is not part of the thundering-herd fix — please drop it for these background hooks, or raise the cap to something like 100 so the effective behavior is "retry forever with ~30s cadence after the first minute". That gives servers breathing room without making the UI fragile.

### 3. `Fixes #2654` is incorrect

Issue #2654 is specifically about **SSE streams** in `hooks/mcp/workloads.ts`, `usePods.ts`, `usePodIssues.ts`, `useDeploymentIssues.ts`, and `useDeployments.ts` — none of which this PR touches. Please either:
- (a) extend the PR to also reconnect the SSE streams named in #2654, or
- (b) remove `Fixes #2654` and retitle this as a WebSocket-only thundering-herd fix, leaving #2654 open.

I'd prefer (b) to keep scope tight.

## Non-blocking

- **Diff noise:** many unrelated cosmetic changes (moving `}` onto its own line inside literal objects, EM-dash to hyphen in a comment in `useUpdateProgress.ts`, `WS_RECONNECT_DELAY` import removal in `useAIPredictions.ts`). These make the diff harder to review — please revert anything that isn't directly part of the backoff fix.
- `getBackoffDelay` is duplicated verbatim in four files. Consider extracting to `lib/wsBackoff.ts` (or similar) so future hooks can import it instead of copy-pasting.
- The `attemptNumber = 0` default parameter on `connect()` in `useClusterProgress` and `useUpdateProgress` is subtle — the ref is the source of truth, and passing `attemptNumber` as an argument creates two paths for the same value. Recommend dropping the parameter and reading `reconnectAttemptsRef.current` directly.

## What's good

- Jitter: correct (`Math.random() * BACKOFF_JITTER_MAX_MS`)
- Max cap: correct (`30_000ms`)
- Reset on `onopen`: correct in all four hooks
- Named constants, no magic numbers
- Unmount guards added to `useUpdateProgress`

Once the giveup behavior is fixed (either removed or cap raised to ~100) and the `Fixes #2654` claim is corrected, this should be good to merge. Thanks again!

khushal-winner:
- Removed max attempt limits for long-lived hooks
- Now retry forever with 30s max delay
- Removed Fixes 2654 from commit title
- Clarified WebSocket-only scope
- Jitter: 0-1000ms prevents synchronized reconnections
- Exponential backoff: 2s → 4s → 8s → 16s → 30s

@clubanderson , is it good to go now?

clubanderson:
Thanks @khushal-winner — this is a solid fix.

Review notes (approving):
- Matches the `useExecSession.ts` backoff pattern (same constant shape, delay values adjusted for hook-level cadence — longer max on 30s vs 16s makes sense for polling hooks vs exec sessions)
- Zero magic numbers — 4 named `WS_RECONNECT_*` / `BACKOFF_JITTER_MAX_MS` / `MAX_WS_RECONNECT_ATTEMPTS` constants with descriptive comments
- Unmount guards + failed-state on retry exhaustion are the right moves
- All 14 blocking CI checks green; 0 failures

One future-follow-up suggestion (not a blocker for this PR): the 4 reconnection constants are duplicated across 4 hooks plus `useExecSession.ts` — worth consolidating into `lib/constants/websocket.ts` in a separate PR once this lands. Happy to file an issue for that.

/lgtm
/approve

Merging via admin-squash.

PR #8342, opened 2026-04-16
Add Playwright test for OAuth vs Demo mode validation

ShaistaAfreen09:
This PR adds a Playwright test to validate authentication behavior in the KubeStellar Console.

The test detects whether the application is running in demo mode or OAuth mode and verifies correct behavior for each case, including OAuth redirect handling.

This contribution focuses on improving test coverage for the authentication path, as highlighted in the mentorship discussion.

### DIFF
--- a/web/e2e/Login.spec.ts
+++ b/web/e2e/Login.spec.ts
@@ -1,7 +1,7 @@
 import { test, expect } from '@playwright/test'
 
 // Login tests require a backend with OAuth enabled.
-// In CI (frontend-only preview builds), /login redirects to the dashboard
+// In CI (frontend only preview builds), /login redirects to the dashboard
 // because there is no auth layer. Skip the whole suite when the backend
 // health endpoint is unreachable.
 test.describe('Login Page', () => {
@@ -17,7 +17,6 @@ test.describe('Login Page', () => {
     await page.goto('/login')
     await page.waitForLoadState('domcontentloaded')
 
-    // Verify login page elements using data-testid
     await expect(page.getByTestId('login-page')).toBeVisible({ timeout: 10000 })
     await expect(page.getByTestId('login-welcome-heading')).toBeVisible()
     await expect(page.getByTestId('github-login-button')).toBeVisible()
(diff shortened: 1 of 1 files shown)


clubanderson:
Thanks for the contribution, @ShaistaAfreen09! Test coverage for the auth path is welcome. A few issues block merge as-is — happy to help iterate:

**Test path doesn't match Playwright config**
- `playwright.config.ts` has `testDir: './e2e'`, so tests need to live in `web/e2e/` (not `web/tests/e2e/`). As written this spec isn't picked up by the runner.

**Duplicates existing `web/e2e/Login.spec.ts`**
- That suite already handles demo-vs-OAuth (probes `/health`, uses `test.skip` when backend is down, validates the login page, and clicks the GitHub button). Worth extending it rather than adding a parallel file — mind checking what's already there first?

**Hardcoded production URL**
- `page.goto('https://console.kubestellar.io/')` bypasses Playwright's `baseURL` (default `http://localhost:8080`, overridable via `PLAYWRIGHT_BASE_URL`). CI runs against the preview / local dev server, so hitting prod:
  1. makes CI depend on production uptime,
  2. pings prod with test traffic,
  3. doesn't actually validate the code under test.
- Use `await page.goto('/')` instead and let the config pick the base URL.

**Selectors**
- Prefer `data-testid` over `:has-text("...")` — text matchers break when i18n translates the button label. The existing Login suite uses `getByTestId('github-login-button')`, `getByTestId('login-page')`, etc.

**Assertion logic**
- The third branch (`else { expect(sidebar-primary-nav visible) }`) assumes "already authenticated", but with `storageState: { cookies: [], origins: [] }` (the pattern in Login.spec.ts) you never reach that state without going through a button, so the branch is effectively dead.
- Missing trailing newline.

Happy to pair on a reshaped version that drops into `web/e2e/Login.spec.ts` as additional test cases — that'd be a clean merge.

ShaistaAfreen09:
@clubanderson Thanks for the detailed feedback ,  this is really helpful.

I will update the PR to align with the existing Playwright setup by:

1. moving the test to the correct directory
2. using the configured baseURL instead of a hardcoded URL
3. switching to data-testid selectors
4. integrating the test into the existing Login.spec.ts suite

I will push an updated version shortly.

ShaistaAfreen09:
Updated the PR based on your feedback:
-  integrated the test into Login.spec.ts
- aligned with Playwright config and baseURL
- switched to data-testid selectors
- removed duplicate test file
- added DCO signoff

Thanks for the guidance , happy to make further improvements if needed.

clubanderson (commented):
Thanks @ShaistaAfreen09! The new `detects demo mode vs OAuth mode behavior` test is a useful smoke check — confirms `/` lands either on login or dashboard depending on mode.

Two nits (non-blocking):

1. **The new test doesn't differentiate the two modes — it asserts "one of these is visible."** The title promises OAuth-vs-demo validation, but both branches are single-expect smokes. For actual mode detection, you'd want to assert `localStorage` or `/api/me` response shape (e.g. `demoMode: true` vs a real user). Happy with the scope as-is if you want to land this and iterate, but worth raising.

2. **You stripped a lot of comments in the existing tests.** The "what does this test do" comments before each `expect` (e.g. `// Verify we're on dashboard (not redirected to login)`) are arguably redundant with the assertion, but some WHY-context comments are load-bearing — e.g. `// Mock MCP endpoints for dashboard`, `// Set localStorage token to simulate authentication`, `// Navigate to home - should stay on dashboard since authenticated`. Those explain **why** the mock/setup exists, which isn't obvious from the next line. Please restore the WHY-context ones (keep stripping the pure restate-the-assertion comments, those are fine to drop).

CI is all green (build × 2, route-smoke, coverage-gate, UI/UX, fullstack-smoke, TTFI, Netlify preview, DCO). Once you address (2), this should be good to merge.

ShaistaAfreen09:
@clubanderson  Thanks! I have restored the WHY context comments as suggested.

Let me know if anything else needs adjustment.

PR #8589, opened 2026-04-17
🐛 fix: Settings page scroll flicker (#8578)

### LINKED ISSUE #8578: bug: Flickering when "System Updates" status is visible
### Describe the bug

When I go to the "Settings" on the left-hand navigation bar and scroll the main area down all the way or nearly all the way, the main area starts flickering rapidly between two different contents, making it hard to read.

The whole second-to-the-left navigation area, for the sections of "Settings", is flickering between two amounts of scrolling. Following are two screen snapshots, showing the two states that the flicker goes between.

Image

Image

### Environment

- Browser: Firefox 149.0.2
- OS: MacOS 15.7.5
- console: 0.3.20

### Steps To Reproduce

See above

### Expected Behavior

Contents readable.

```shell

```

### Want to contribute?

### Additional Context

_No response_

clubanderson:
## Summary

Fixes #8578

When scrolled to the bottom of the Settings page, the `IntersectionObserver` oscillated between two active sections. Each `setActiveSection` call triggered a re-render that shifted the sidebar's scroll position, creating a rapid flicker loop — particularly visible in Firefox.

## Root Cause

The observer uses `rootMargin: '0px 0px -40% 0px'` to shrink its observation zone to the top 60% of the viewport. At the bottom of the page, sections near the boundary oscillate between intersecting and not-intersecting states. Each state change re-renders the sidebar nav (different item gets the active highlight), which causes a micro layout shift that re-triggers the observer — creating a feedback loop.

## Fix

1. **Bottom-of-scroll stabilization**: When the scroll container is within 50px of its maximum scroll position, pick the **last** visible section instead of the first. The last section stays consistently visible at the bottom boundary, preventing oscillation.

2. **Skip no-op re-renders**: Track `activeSection` in a ref alongside the state. The observer callback checks the ref before calling `setActiveSection`, skipping the call entirely when the value hasn't changed. This eliminates unnecessary re-renders that feed the layout loop.

## Test plan

- [ ] Open Settings page, scroll to the very bottom — sidebar should not flicker
- [ ] Firefox specifically (where the issue was reported)
- [ ] Clicking sidebar nav items still scrolls to the correct section
- [ ] Deep linking via URL hash (e.g. `/settings#system-updates-settings`) still works
- [ ] IntersectionObserver still highlights the correct section during normal scrolling

### DIFF
--- a/web/src/components/settings/Settings.tsx
+++ b/web/src/components/settings/Settings.tsx
@@ -100,6 +100,10 @@ export function Settings() {
   const { restoredFromFile, syncStatus, lastSaved, filePath, exportSettings, importSettings } = usePersistedSettings()
 
   const [activeSection, setActiveSection] = useState<string>('ai-mode-settings')
+  // Mirror of activeSection for the IntersectionObserver callback — avoids
+  // reading stale closure values and lets us skip no-op setState calls that
+  // would trigger re-renders and feed the layout loop (issue #8578).
+  const activeSectionRef = useRef<string>('ai-mode-settings')
   const [showRestoredToast, setShowRestoredToast] = useState(false)
 
   /** Close settings and return to previous page (or home as fallback).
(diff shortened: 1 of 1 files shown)


clubanderson:
Closing — a fix was already merged in #8587 (commit 1a111329). That PR uses a debounce + single threshold approach. Both fixes address the same root cause (IntersectionObserver feedback loop).

PR #8947, opened 2026-04-19
fix: hide demo nodes in Nodes stat block drill-down when in live mode

### LINKED ISSUE #8840: Unable to see nodes in the Dashboard Nodes stat block drillDown
## User Request

**Type:** bug
**Submitted by:** @xonas1101
**Console Request ID:** acd7e01d-c311-44cf-9136-6807ac94a1c2

## Description

On clicking the nodes stat block in the landing dashboard, we fall back to demo data, not seeing the actual nodes list. We can see the nodes of each cluster respectively in the cluster info crds drillDown on the my clusters dashboard though. We should be able to see the same list cumulated across clusters.

---

clubanderson:
## Summary

- The Nodes stat block on the landing Dashboard could open a drill-down filled with 4 placeholder demo nodes (`prod-east`/`vllm-d`/`kind-local`) even when the user was NOT in demo mode.
- Root cause: `useCachedNodes()` uses `demoWhenEmpty: true`. When the `/api/mcp/nodes` fan-out returns empty across all clusters (typically because the caller lacks list-nodes RBAC), it falls back to the demo dataset and the drill-down silently renders those demo nodes.
- Fix: in `MultiClusterSummaryDrillDown`, treat the demo-fallback payload as empty when `isDemoMode` is false. This surfaces the existing empty-state message ("Cluster summary reports N nodes, but the detailed list is empty") that reports the expected count from the per-cluster summary and explains the likely RBAC / endpoint failure — matching what the per-cluster nodes drill-down already shows on My Clusters.
- The 'Demo' freshness badge is also gated on `isDemoMode` so it doesn't appear in live mode.

## Test plan

- [ ] In live mode on a cluster where the current user cannot list nodes, click the Nodes stat block on the landing dashboard → the drill-down should show the empty-state card with the expected count, not the 4 demo placeholders.
- [ ] In live mode with real node data, the drill-down still shows the live nodes aggregated across clusters.
- [ ] In demo mode, the drill-down still shows the demo nodes with the 'Demo' badge.
- [ ] The per-cluster Cluster Info drill-down on My Clusters continues to work (unchanged).

Fixes #8840

### DIFF
--- a/web/src/components/drilldown/views/MultiClusterSummaryDrillDown.tsx
+++ b/web/src/components/drilldown/views/MultiClusterSummaryDrillDown.tsx
@@ -4,6 +4,7 @@ import { useClusterData } from '../../../hooks/useClusterData'
 import { useDrillDownActions } from '../../../hooks/useDrillDown'
 import type { DrillDownViewType } from '../../../hooks/useDrillDown'
 import { useCachedNodes, useCachedPVCs } from '../../../hooks/useCachedData'
+import { useDemoMode } from '../../../hooks/useDemoMode'
 import { useTranslation } from 'react-i18next'
 import { formatRelativeTime } from '../../../lib/formatters'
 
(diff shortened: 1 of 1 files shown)


clubanderson:
Superseded by #8959 which does proper cross-cluster aggregation (DeduplicatedClusters rule per MEMORY.md), addressing the full user request for a cumulative nodes list rather than just suppressing the demo fallback.

PR #9022, opened 2026-04-19
🐛 : Fix path traversal vulnerability in GitOps handlers

khushal-winner:
## Summary

Add validation for 
eq.Path parameter to prevent path traversal attacks via ../ sequences. The vulnerability allowed checking files outside the temp directory in isKustomizeDir function.

## Changes

- Add alidateGitopsPath in pkg/agent/server_gitops.go
- Add alidatePath in pkg/api/handlers/gitops.go
- Call validation in handleDetectDrift and handleGitopsSync (agent)
- Call validation in detectDriftViaKubectl and syncViaKubectl (backend)
- Blocks .., null bytes, and dangerous characters
- Follows existing validation patterns (alidateBranchName)

## Security Impact

Prevents path traversal attacks that could allow checking file existence outside the intended temp directory. The fix is applied to both agent and backend handlers for defense-in-depth.

## Testing

- Both packages build successfully
- Validation is called before path construction in all 4 vulnerable locations
- No breaking changes to existing functionality

### DIFF
--- a/pkg/agent/server_gitops.go
+++ b/pkg/agent/server_gitops.go
@@ -120,6 +120,35 @@ func validateGitopsBranchName(branch string) error {
 	return nil
 }
 
+// validateGitopsPath validates a repository path parameter.
+// SECURITY: Prevents path traversal attacks and flag injection.
+func validateGitopsPath(path string) error {
+	if path == "" {
+		return nil // Empty path is OK - refers to repo root
+	}
+	// Block null bytes
+	if strings.ContainsRune(path, 0) {
+		return fmt.Errorf("path contains null bytes")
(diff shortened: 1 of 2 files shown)


clubanderson:
Thanks for the defense-in-depth fix @khushal-winner! The `verify / verify` check is failing because our PR title convention requires a leading emoji. For a security fix like this, either of these would pass:

- `⚠️ Fix path traversal vulnerability in GitOps handlers`
- `🐛 Fix path traversal vulnerability in GitOps handlers`

You can rename via the edit-title action in the PR header, and the verifier will re-run automatically. No rebase needed.

PR #9026, opened 2026-04-19
Add filepath.Clean to isKustomizeDir to prevent path traversal

khushal-winner:
## Feature Description
Add path normalization to prevent potential path traversal in shell command execution.

## Problem
The isKustomizeDir function receives user-controlled paths via req.Path but lacks normalization before executing shell commands. The codebase already acknowledges path traversal as a security threat (see cleanupTempDir which explicitly checks for '..').

## Solution
Add cleanPath := filepath.Clean(path) before constructing shell command paths.

## Changes
- Add filepath import
- Add cleanPath := filepath.Clean(path) before constructing shell command paths
- Use cleanPath instead of path in exec.Command calls

## Security Value
Defense-in-depth security measure consistent with existing validation patterns in the codebase. Zero performance impact (string operation only) and follows standard Go practice for path handling.

## Testing
Build passed: \go build ./pkg/api/handlers/...\ succeeded.

### DIFF
--- a/pkg/agent/server_gitops.go
+++ b/pkg/agent/server_gitops.go
@@ -120,6 +120,35 @@ func validateGitopsBranchName(branch string) error {
 	return nil
 }
 
+// validateGitopsPath validates a repository path parameter.
+// SECURITY: Prevents path traversal attacks and flag injection.
+func validateGitopsPath(path string) error {
+	if path == "" {
+		return nil // Empty path is OK - refers to repo root
+	}
+	// Block null bytes
+	if strings.ContainsRune(path, 0) {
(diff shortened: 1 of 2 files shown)


clubanderson:
Thanks for the quick follow-up! One small fix needed for CI: the `verify / verify` check requires an emoji prefix on the PR title (the repo convention is ⚠️/✨/🐛/📖/🚀/🌱). Could you edit the title to something like `🔒 Add filepath.Clean to isKustomizeDir to prevent path traversal` or `🐛 fix(gitops): ...`? Once that's done and the remaining builds settle, we can merge.

PR #9351, opened 2026-04-21
test(hooks): fix waitFor boolean return anti-pattern

clubanderson:
## Summary

- `waitFor(fn)` in `@testing-library/react` **only retries when the callback throws**. Returning `false` resolves immediately, leaving state stale and causing assertions to fail.
- Converts all boolean-expression `waitFor` callbacks to `expect()`-based assertions across three test files: `useBonusPoints`, `useServiceImportsCard`, `useGatewayStatus`.

**Root cause example:**
```ts
// Before — resolves immediately when bonusPoints is 0 (returns false, no throw)
await waitFor(() => result.current.bonusPoints === 150)

// After — retries until bonusPoints reaches 150 (throws AssertionError until then)
await waitFor(() => expect(result.current.bonusPoints).toBe(150))
```

**Evidence:** `useServiceImportsCard` test suite completed in 97ms for 22 tests — far less than the 1000ms `waitFor` default timeout, confirming all `waitFor` calls were resolving immediately.

## Test plan
- [ ] Shards 6 and 8 pass in CI (no more `expected +0 to be 150` or `expected [] to have a length of 1`)
- [ ] All 3 affected test files pass locally

### DIFF
--- a/web/src/hooks/__tests__/useBonusPoints.test.ts
+++ b/web/src/hooks/__tests__/useBonusPoints.test.ts
@@ -101,8 +101,7 @@ describe('useBonusPoints — successful fetch', () => {
     authenticatedUser('bob')
     mockFetchBonus({ login: 'bob', total_bonus_points: 150, entries: [] })
     const { result } = renderHook(() => useBonusPoints())
-    await waitFor(() => result.current.bonusPoints === 150)
-    expect(result.current.bonusPoints).toBe(150)
+    await waitFor(() => expect(result.current.bonusPoints).toBe(150))
   })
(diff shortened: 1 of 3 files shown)


clubanderson:
Merge conflict detected against `main` — PR #9350 landed with overlapping waitFor fixes. Rebase needed before admin-merge. Skipping for this iteration; another session's agent can reconcile.

clubanderson:
Superseded by #9350 which merged the same fixes. This PR's diff is now a subset — remaining work will be a new targeted patch.

Issue #10162, opened 2026-04-26
bug: ImportDashboard endpoint bypasses per-user dashboard limit


Va16hav07:
The `POST /api/dashboards/import` endpoint (`ImportDashboard` in `pkg/api/handlers/dashboard.go`) creates a new dashboard directly via `h.store.CreateDashboard()` **without** first checking whether the user has already reached the `MaxDashboardsPerUser` (50) limit. The `CreateDashboard` handler enforces this limit correctly, but `ImportDashboard` bypasses it entirely, allowing a user to import unlimited dashboards by calling the import endpoint in a loop.

**What happened:** A user can send repeated `POST /api/dashboards/import` requests and accumulate more than 50 dashboards, exceeding the `MaxDashboardsPerUser = 50` limit defined in `dashboard.go`. No error is returned.

**What was expected:** The import endpoint should return HTTP 429 (Too Many Requests) with a clear message once the user already has 50 dashboards, consistent with the behaviour of the regular `CreateDashboard` endpoint.

### Environment

- Browser: (e.g., Chrome 120, Firefox 121)
- OS: (e.g., Linux, macOS 14.0)
- KC Version: (e.g., 0.1.0)
- Kubernetes cluster type: (e.g., kind, EKS)

### Steps To Reproduce

1. Create 50 dashboards for a user via `POST /api/dashboards` until the 429 limit is hit.
2. Switch to the import endpoint: `POST /api/dashboards/import` with a valid `kc-dashboard-v1` payload.
3. Observe that the import succeeds even though the user is already at the dashboard limit.
4. Repeat step 2 — each call creates an additional dashboard beyond the 50-dashboard cap.

### Expected Behavior

`POST /api/dashboards/import` should call `h.store.CountUserDashboards()` before `h.store.CreateDashboard()`, exactly as `CreateDashboard` does, and return HTTP 429 when the count is at or above `MaxDashboardsPerUser`.

```go
// Should be added to ImportDashboard before h.store.CreateDashboard():
count, err := h.store.CountUserDashboards(c.UserContext(), userID)
if err != nil {
    return fiber.NewError(fiber.StatusInternalServerError, "Failed to check dashboard count")
}
if count >= MaxDashboardsPerUser {
    return fiber.NewError(fiber.StatusTooManyRequests,
        fmt.Sprintf("Dashboard limit reached (%d), maximum is %d per user", count, MaxDashboardsPerUser))
}
```

### Browser Console / Backend Logs

```shell

```

[...]

PR #10166, opened 2026-04-26
fix: enforce per-user dashboard limit in ImportDashboard (#10162)

vedparkasharya:
## Problem

The `POST /api/dashboards/import` endpoint (`ImportDashboard`) created dashboards directly via `h.store.CreateDashboard()` **without** checking whether the user had reached the `MaxDashboardsPerUser` (50) limit. This allowed users to bypass the dashboard cap by repeatedly calling the import endpoint.

The `CreateDashboard` handler correctly enforces this limit, but `ImportDashboard` bypassed it entirely.

## Solution

Added the same `CountUserDashboards` check that `CreateDashboard` uses, placed **before** any dashboard creation or card validation:

```go
count, err := h.store.CountUserDashboards(c.UserContext(), userID)
if err != nil {
    return fiber.NewError(fiber.StatusInternalServerError, "Failed to check dashboard count")
}
if count >= MaxDashboardsPerUser {
    return fiber.NewError(fiber.StatusTooManyRequests,
        fmt.Sprintf("Dashboard limit reached (%d), maximum is %d per user", count, MaxDashboardsPerUser))
}
```

## Changes

- `pkg/api/handlers/dashboard.go`: Added per-user dashboard limit check in `ImportDashboard`
- `pkg/api/handlers/dashboard_test.go`: Added `TestImportDashboard_ExceedsUserLimit` to verify HTTP 429 is returned when limit is reached

## Testing

- Added `TestImportDashboard_ExceedsUserLimit` that mocks `CountUserDashboards` returning `MaxDashboardsPerUser` and asserts HTTP 429
- Existing `TestImportDashboard_Success` continues to pass (default mock returns 0, under limit)

## Checklist

- [x] Bug fix (non-breaking change)
- [x] Added tests
- [x] Follows existing code style
- [x] References issue #10162

## Proof

The fix mirrors the exact pattern already used in `CreateDashboard` (~line 106), ensuring consistent behavior across both endpoints.

### DIFF
--- a/pkg/api/handlers/dashboard.go
+++ b/pkg/api/handlers/dashboard.go
@@ -271,6 +271,16 @@ func (h *DashboardHandler) ImportDashboard(c *fiber.Ctx) error {
 		input.Name = "Imported Dashboard"
 	}
 
+	// Enforce per-user dashboard limit (#10162) — same check as CreateDashboard.
+	count, err := h.store.CountUserDashboards(c.UserContext(), userID)
+	if err != nil {
+		return fiber.NewError(fiber.StatusInternalServerError, "Failed to check dashboard count")
+	}
+	if count >= MaxDashboardsPerUser {
+		return fiber.NewError(fiber.StatusTooManyRequests,
+			fmt.Sprintf("Dashboard limit reached (%d), maximum is %d per user", count, MaxDashboardsPerUser))
+	}
+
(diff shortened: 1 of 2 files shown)


clubanderson (approved):
LGTM — clean fix that mirrors the existing `CreateDashboard` limit check. Test coverage is correct. Nice first contribution! 🎉

PR #10368, opened 2026-04-27
security: mask agent token + validate proxy subPath (#10363 follow-up)

clubanderson:
## Summary
- Stop echoing KC_AGENT_TOKEN value to stdout in startup-oauth.sh (prevents terminal/log leaks)
- Add path validation to kc-agent proxy handler — restrict subPath to allowlisted endpoints (`status`, `config`, `trigger`, `cancel`), reject traversal sequences

Fixes Copilot review comments on #10363.

## Test plan
- [x] startup-oauth.sh no longer prints token value
- [x] Proxy rejects `..` and non-allowlisted paths with 400

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -1054,9 +1054,23 @@ func (s *Server) setupRoutes() {
 	// and Private Network Access checks that are fragile across browser versions.
 	// The backend injects the Bearer token server-side so the frontend never
 	// needs to know KC_AGENT_TOKEN for auto-update operations.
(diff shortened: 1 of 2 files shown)


clubanderson (commented):
Review: LGTM — token masking and allowlist-based path validation look correct.

Minor defense-in-depth suggestion: consider URL-decoding `subPath` after extraction and re-checking for traversal sequences (e.g., double-encoded `%252e`). Fiber likely pre-decodes path parameters, so this is belt-and-suspenders, not blocking.

Overall: solid security improvement. ✅

Issue #10581, opened 2026-04-27
The TypeScript configuration (`tsconfig.json`) contains an invalid value


arnavgogia20:
## User Request

**Type:** bug
**Target:** Console Application
**Submitted by:** @arnavgogia20
**Console Request ID:** 533851aa-8a08-4a30-acb6-41db079b91d5

### Summary
The TypeScript configuration (`tsconfig.json`) contains an invalid value for the `ignoreDeprecations` compiler option, causing a TypeScript validation error in the editor.

---

### Location
console/web/tsconfig.json  

---

### Error
Invalid value for '--ignoreDeprecations'. ts  

---

### Steps to Reproduce
Open the project in VS Code  
Navigate to console/web/tsconfig.json  
Observe the `compilerOptions` section  
Notice the following configuration:  

```
"ignoreDeprecations": "6.0"
```

VS Code/TypeScript flags this as an invalid value  

---

### Expected Behaviour
The `tsconfig.json` should not produce any TypeScript validation errors  

All compiler options should use valid and supported values  

---

### Actual Behaviour
TypeScript reports:  
Invalid value for '--ignoreDeprecations'  

The config is marked with an error in the editor  

---

### Possible Causes
The value `"6.0"` is not a valid accepted value for `ignoreDeprecations`  
Mismatch between TypeScript version and supported config options  
Option may require a boolean or a different format depending on TS version  

---

### Suggested Fix
Update or remove the `ignoreDeprecations` field  

Possible fixes:  
Use a valid supported value based on the current TypeScript version  
Upgrade/downgrade TypeScript to match expected config behaviour  
Remove the option if not required  

---

### Impact
Causes editor/type-checking errors  
Confusing for contributors setting up the project  
May affect build or tooling consistency  

---

---
*This issue was automatically created from the KubeStellar Console.*

clubanderson:
Closing: `ignoreDeprecations: "6.0"` is correct and required for TypeScript 6.0.3 (our current version). It suppresses the TS 6.0 deprecation warning for `baseUrl`, which we need for path aliases. Removing it causes a build error:

```
error TS5101: Option 'baseUrl' is deprecated and will stop functioning in TypeScript 7.0.
Specify compilerOption '"ignoreDeprecations": "6.0"' to silence this error.
```

The VS Code error is likely caused by an older TypeScript version in the editor's language service. Running `npx tsc --version` in the project should show 6.0.3.

PR #10665, opened 2026-04-28
fix(tsconfig): downgrade ignoreDeprecations from "6.0" to "5.0" for compatibility

arnavgogia20:
### Summary

Downgrades the `ignoreDeprecations` value in `tsconfig.json` from `"6.0"` to `"5.0"` to ensure compatibility with the current TypeScript version used in the project.

---

### Problem

The project currently specifies:

```
"ignoreDeprecations": "6.0"
```

This value is not supported by the TypeScript version used by the project/tooling, resulting in the following error during development:

```
Invalid value for '--ignoreDeprecations'
```

This breaks local setup and prevents contributors from compiling or running the project successfully.

---

### Solution

Updated the configuration to:

```
"ignoreDeprecations": "5.0"
```

This ensures compatibility with the existing TypeScript setup while still preserving the intent of suppressing deprecation warnings.

---

### Why this approach

* Maintains the purpose of the `ignoreDeprecations` setting instead of removing it entirely
* Avoids forcing a full TypeScript version upgrade across the codebase
* Ensures smooth onboarding and local development for contributors

---

### Impact

* Fixes TypeScript configuration error
* Restores working local development environment
* No changes to runtime or application behavior

---

### Testing

* Ran `npm install` and verified setup
* Confirmed TypeScript no longer throws configuration error
* Project runs/builds successfully after change

---

### Note

If/when the project upgrades to a TypeScript version that supports `"6.0"`, this value can be safely reverted.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "kubestellar-console-status-helpers",
+  "name": "console",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {}
--- a/web/tsconfig.json
+++ b/web/tsconfig.json
@@ -16,7 +16,7 @@
     "noUnusedParameters": true,
     "noFallthroughCasesInSwitch": true,
     "noUncheckedSideEffectImports": true,
-    "ignoreDeprecations": "6.0",
+    "ignoreDeprecations": "5.0",
     "baseUrl": ".",
     "paths": {
       "@/*": ["src/*"]


arnavgogia20:
@clubanderson Noticed that the failing checks seem unrelated to this change (tsconfig update only).
The failures appear to be in build/deploy and performance pipelines..... possibly due to upstream or environment issues.
Happy to investigate further if needed, but wanted to confirm if this is expected or if CI is currently unstable.

I had also created an issue regarding this earlier ( #10581 ).....and thought it was a small change so i opened a PR for it

PR #11138, opened 2026-04-30
Fix settings save failure after kc-agent restart

clubanderson:
## Summary
- **Settings "Save failed" bug**: `agentFetch` now detects 401 from a stale cached kc-agent token, clears it from localStorage, fetches a fresh token from `/api/agent/token`, and retries — fixing the persistent "Save failed" state after kc-agent restarts
- **Error capture limits**: Increased browser error ring buffer from 30→50 entries and GitHub issue display cap from 20→50 so all captured errors are shown without truncation

## Test plan
- [ ] Restart kc-agent (via `startup-oauth.sh`), verify settings save works without manual page reload
- [ ] Submit a bug report with >20 console errors, verify all are shown in the issue body
- [ ] Verify normal settings save/restore still works when token is valid


clubanderson:
Closing — duplicate of #11137 which has been merged.

PR #11159, opened 2026-04-30
🐛 Fix Worker mock leak in cache-coverage.test.ts

### LINKED ISSUE #11150: 🐛 Worker mock leak in cache-coverage.test.ts — globalThis.Worker not restored in finally block
## Bug Report

**Source:** Copilot review on PR #11039 (merged 2026-04-30)
**Severity:** High — can cause flaky tests and false failures

## Root Cause

In `web/src/lib/cache/__tests__/cache-coverage.test.ts:1053`, `globalThis.Worker` is reassigned for the test but only restored at the end of the happy path. If an assertion throws (or the test times out), the Worker mock leaks into subsequent tests and can cause spurious failures.

## Fix Required

Wrap the Worker reassignment in a `try/finally` block to guarantee restoration:

```typescript
const originalWorker = globalThis.Worker;
try {
  globalThis.Worker = MockWorker as any;
  // ... test assertions ...
} finally {
  globalThis.Worker = originalWorker;
}
```

Or use `beforeEach`/`afterEach` hooks to save and restore the mock.

## File

`web/src/lib/cache/__tests__/cache-coverage.test.ts` around line 1053

**Address Copilot review findings from PR #11039**

clubanderson:
Wraps `globalThis.Worker` reassignment in `try/finally` to prevent mock leak when assertions throw.

Previously the describe-level `originalWorker` variable was restored via `afterEach` only when truthy, leaving a window where a falsy `Worker` (e.g. `undefined` in jsdom) would not be restored. The `try/finally` guarantees cleanup unconditionally.

Changes:
- Move `originalWorker` into a `const` local to the test body
- Wrap all Worker-dependent code in `try/finally` for guaranteed cleanup
- Simplify `afterEach` to only reset modules (Worker restore is now in-test)

Fixes #11150

Copilot review finding from PR #11039.

### DIFF
--- a/web/src/lib/cache/__tests__/cache-coverage.test.ts
+++ b/web/src/lib/cache/__tests__/cache-coverage.test.ts
@@ -1004,69 +1004,68 @@ describe('saveMeta localStorage fallback', () => {
 // ============================================================================
 
 describe('worker-active IndexedDB mirror write', () => {
-  let originalWorker: typeof Worker
-
   afterEach(() => {
-    // Always restore Worker and reset modules to prevent mock leakage into subsequent tests.
-    // This is critical because globalThis.Worker is a global singleton — if not restored,
-    // subsequent tests will get the mocked version. See #11039.
-    if (originalWorker) {
-      globalThis.Worker = originalWorker
-    }
+    // Reset modules to prevent mock leakage into subsequent tests. See #11039.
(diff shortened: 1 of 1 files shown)


clubanderson:
Superseded by #11157 which was merged first with the same fix (Worker mock try/finally cleanup).

PR #11242, opened 2026-05-01
✨ Quick fixes: demo bypass, HTTP timeout, resource cleanup, console restore, i18n, AbortSignal timeout

clubanderson:
Fixes bundled from architect and reviewer passes:

- ACMMBadgeHandler demo mode bypass (prevents real data leak in demo)
- HTTP fallback client 10s timeout (prevents connection hangs)  
- missions.go resp.Body.Close() → defer (prevents fd leaks)
- analytics-core.ts console monkey-patch cleanup on re-init
- Missions.tsx i18n wrapping + 20 locale entries
- GitHubActivity.tsx AbortSignal.timeout on all unguarded fetch calls

### DIFF
--- a/pkg/api/handlers/acmm_badge.go
+++ b/pkg/api/handlers/acmm_badge.go
@@ -147,7 +147,7 @@ func ACMMBadgeHandler(c *fiber.Ctx) error {
 			Color:         "blue",
 			CacheSeconds:  3600,
 		}
-		return serveBadge(c, badge, 3600)
+		return serveBadge(c, badge, badgeCacheControlMaxAge)
 	}
 
 	// Look up cache
--- a/pkg/api/handlers/acmm_badge.go
+++ b/pkg/api/handlers/acmm_badge.go
@@ -138,6 +138,18 @@ func ACMMBadgeHandler(c *fiber.Ctx) error {
 		}, badgeCacheControlErrorSecs)
 	}
 
+	// Demo mode support
+	if isDemoMode(c) {
+		badge := &shieldsEndpointBadge{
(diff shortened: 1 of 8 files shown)


clubanderson (commented):
## Code Review

### 🔴 Critical: Type mismatch in `serveBadge` call
**File:** `pkg/api/handlers/acmm_badge.go:150`

`serveBadge(c, badge, 3600)` passes an `int` where a `string` is expected. This causes a Go compilation error:
```
cannot use 3600 (untyped int constant) as string value in argument to serveBadge
```

**Fix:** Use `badgeCacheControlMaxAge` (the string constant used by all other callers) instead of `3600`.

---

All other changes (defer resp.Body.Close, HTTP 10s timeout, console cleanup, i18n, AbortSignal) look correct and well-motivated.

clubanderson:
🤖 **Reviewer**: Addressed MEDIUM Copilot comments (commit `6b399b02f`):

**Missions.tsx i18n namespace syntax (lines 123, 148)**:
- Fixed all `t('cards.X')` → `t('cards:X')` and `t('common.X')` → `t('common:X')` — with `useTranslation(['common','cards'])`, dot-notation was resolving keys in the `common` namespace, causing non-English locales to always fall back to the hardcoded English string
- Also fixed `t('common.searchMissions')` → `t('cards:searchMissions')` (key lives in cards.json)

**missions.go `fetchWithCache` defer-in-loop (line 532)**:
- Replaced `defer resp.Body.Close()` inside the retry loop with an explicit close immediately after `io.ReadAll` — deferring inside a loop keeps multiple response bodies open until the function returns

PR #11243, opened 2026-05-01
🐛 fix kagent-crds agent struct, jaeger metrics guard, liveMocks shapes

clubanderson:
Fixes two source-file bugs that caused `[DynamicCard:KagentAgentFleet] Render error: React error #31` and `TypeError: Cannot read properties of undefined (reading 'servicesCount')` in nightly Playwright runs.

### 1. `pkg/agent/kagent_crds.go` — wrong JSON field names
The `kagentCRDAgent` struct used `Type json:"type"` (TS expects `agentType`) and two `bool` fields `Ready`/`Accepted` (TS expects a single `status string`). In production every agent showed an empty/broken status badge. In tests with the mock data the mismatch caused an object to be passed where a string was expected, triggering React error #31.

**Fix:** Rename field to `AgentType json:"agentType"`, replace bool fields with `Status string json:"status"` populated by `extractConditionStatus()`. Add `Status` field to `kagentCRDTool`, `kagentCRDModel`, `kagentCRDMemory` via new `extractReadyStatus()` helper.

### 2. `web/src/components/cards/jaeger_status/index.tsx` — unguarded `data.metrics`
Line 84: `data.metrics.servicesCount > 0` crashes when the API/mock returns a response without the `metrics` sub-object. `INITIAL_DATA` has it but a stale or bad cache entry does not.

**Fix:** Optional-chain the `hasAnyData` check; destructure a fallback `const metrics` at render time.

### 3. `web/e2e/mocks/liveMocks.ts` — raw K8s CRD shapes in mock data
`kagent-crds/*` entries had raw K8s `{ kind, apiVersion, spec: {...}, status: { phase: 'Ready' } }` objects instead of the flat backend-processed shapes the TypeScript types expect. `status: { phase: 'Ready' }` (object) rendered as a JSX child → React error #31 crash.
`jaeger-status` stub missing `metrics`, `collectors`, `query`, `version` fields.

**Fix:** Replace all four `kagent-crds/*` entries with flat `KagentCRDAgent/Tool/Model/Memory`-shaped objects. Replace jaeger stub with full `JaegerStatus`-shaped object.

## Testing
- `/usr/local/go/bin/go build ./pkg/agent/...` ✅
- `cd web && npm run build` ✅
- `cd web && npm run lint` ✅ (exit 0)

### DIFF
--- a/pkg/agent/kagent_crds.go
+++ b/pkg/agent/kagent_crds.go
@@ -57,15 +57,15 @@ var (
 	}
 )
 
-// kagentCRDAgent is the JSON response shape for a kagent.dev Agent CR
+// kagentCRDAgent is the JSON response shape for a kagent.dev Agent CR.
+// Field names match the TypeScript KagentCRDAgent interface.
 type kagentCRDAgent struct {
 	Name           string `json:"name"`
 	Namespace      string `json:"namespace"`
 	Cluster        string `json:"cluster"`
-	Type           string `json:"type"`
+	AgentType      string `json:"agentType"`
 	Runtime        string `json:"runtime"`
-	Ready          bool   `json:"ready"`
-	Accepted       bool   `json:"accepted"`
+	Status         string `json:"status"`
 	ModelConfigRef string `json:"modelConfigRef"`
 	ToolCount      int    `json:"toolCount"`
 }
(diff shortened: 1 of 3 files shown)


clubanderson (commented):
## Code Review

### 🔴 Critical: Missing fields in `kagentCRDAgent` Go struct
**File:** `pkg/agent/kagent_crds.go`

The struct is missing fields the TypeScript UI actively uses:
- `replicas` / `readyReplicas` — used in KagentAgentFleet.tsx to render replica counts
- `a2aEnabled` — used in KagentAgentDiscovery.tsx to filter and badge A2A agents
- `systemMessage`, `createdAt`, `age` — defined in TS interface

Without these, the UI gets `undefined` and renders incorrectly.

### 🟡 Medium: Invalid runtime value in liveMocks
**File:** `web/e2e/mocks/liveMocks.ts`

`runtime: 'googleadk'` doesn't match the TypeScript union `'python' | 'go' | ''`. Either update the mock to use a valid value or extend the TS type.

### 🔴 DCO: Commits missing `Signed-off-by`
CI shows DCO failure. All commits need `git commit -s` sign-off.

clubanderson:
🐛 Fixed Go build failure: `pkg/api/handlers/acmm_badge.go:150` was passing literal `3600` (int) as the third argument to `serveBadge()`, which expects a `string` cache-control header. Changed to use named constant `badgeCacheControlMaxAge` (same pattern as all other call-sites). Pushed as e4204103c — CI re-running.

PR #11278, opened 2026-05-01
🐛 Fix nightly validation pass rate check grep pattern mismatch

### LINKED ISSUE #11246: 🐛 bug: Nightly validation pass rate check never executes due to grep pattern mismatch
### What happened?
The KB nightly validation workflow has a regression check that's supposed to fail CI when the mission pass rate drops below 90%. However, this check is **dead code** that never executes.

The workflow step (`.github/workflows/kb-nightly-validation.yml`, lines 26-34) searches for the literal string `"pass rate"`:

```yaml
if grep -q "pass rate" validation-results.txt; then
  RATE=$(grep -oP '\d+%' validation-results.txt | head -1 | tr -d '%')
  if [ "${RATE:-100}" -lt 90 ]; then
    echo "::error::Mission validation pass rate dropped below 90%"
    exit 1
  fi
fi
```

But the validation script (`scripts/validate-missions.sh`, line ~350) never writes that string. It actually outputs:

```bash
echo -e "  Passed (no critical issues): ${GREEN}${passed}${RESET} (${pass_pct}%)"
```

**Result**: `grep -q "pass rate"` returns non-zero (string not found) → entire `if` block is skipped → workflow always exits 0 regardless of actual pass rate.

### What did you expect to happen?
The workflow should:
1. **Parse the actual pass rate** from validation output (e.g., 40%, 85%, 98%)
2. **Fail CI when pass rate < 90%** as documented in the workflow comment
3. **Hard-fail if pass rate cannot be parsed** (no silent skip)

This would prevent degraded KB content (stale runbooks, broken chart references, version drift) from being marked as "validated" and served to the AI assistant.

### Impact
This causes **silent KB quality degradation**:

- **CI shows green** even when KB pass rate drops from 98% → 40%
- **AI assistant sees `qualityPass: true`** for stale content
- **SREs trust AI recommendations** backed by false validation signals
- **Clusters fail to reconcile** when executing outdated runbooks (renamed dependencies, wrong apiVersions)
- **No alerts fired** - the only enforcement mechanism is dead code

**LFX Mentorship Impact**: 
The entire project is measured on "validated KB" and "reliable mission control pass rates" - this bug means those metrics have **no CI enforcement**, undermining the core deliverable.

### Root Cause
1. **String mismatch**: Producer writes `"Passed (no critical issues)"`, consumer greps for `"pass rate"`
2. **No test coverage**: The regression-check failure path has never been exercised in CI
3. **Fragile prose-grep**: Both producer and consumer rely on substring matching instead of machine-parseable output

The script was likely reworded at some point, or the workflow was written against planned output that never landed. Since there's no test that validates the gate actually fires, the dead code has been silent since merge.

### Environment

**Affected Components**:
- CI/CD: GitHub Actions workflow `.github/workflows/kb-nightly-validation.yml`
- Validation Script: `scripts/validate-missions.sh`
- KB Repository: `kubestellar/console-kb`

**Not environment-dependent** - this is pure CI/CD logic failure:
- No browser required
- No Kubernetes cluster required  
- No KC version dependency
- Reproducible with `bash` and `grep` alone

**Trigger**: 
Every nightly cron run at 04:00 UTC (or any manual workflow dispatch)

### Reproduction (No Cluster Needed)

**1. Simulate KB degradation**:
```bash
cd console-kb
# Remove validated_at from 50 missions (creates warning-class issues)
find fixes/ -name "*.yaml" -type f | head -50 | xargs sed -i '/validated_at:/d'
```

**2. Run validation script**:
```bash
./scripts/validate-missions.sh > validation-results.txt
echo $?  # Exits 0 (critical == 0, warnings don't fail)
```

**3. Check actual output**:
```bash
cat validation-results.txt
# Shows: "Passed (no critical issues): 120 (40%)"
#        ^^^^^^^^^^^^^^^^^^^^^^^^^^
#        This string is what's actually written
```

**4. Test workflow gate**:
```bash
# This is what the workflow does
grep -q "pass rate" validation-results.txt
echo $?  # Returns 1 (string not found)

# Gate is skipped, so threshold check never runs
if grep -q "pass rate" validation-results.txt; then
  echo "This block never executes"
fi
```

**5. Verify workflow exits 0**:
```bash
# Simulate full workflow step
if grep -q "pass rate" validation-results.txt; then
  RATE=$(grep -oP '\d+%' validation-results.txt | head -1 | tr -d '%')
  if [ "${RATE:-100}" -lt 90 ]; then
    exit 1
  fi
fi
echo "Workflow exits: $?"  # Always 0, even with 40% pass rate
```

### Expected Behavior

```bash
# Workflow should fail when pass rate < 90%
RATE=40
if [ "${RATE}" -lt 90 ]; then
  echo "::error::Mission validation pass rate dropped below 90% (currently ${RATE}%)"
  exit 1  # ← This should happen but never does
fi
```

```shell

```

### Want to contribute?

### Additional Context

_No response_

clubanderson:
Fixes #11246

## Problem

The KB nightly validation workflow grepped for `"pass rate"` in `validation-results.txt`, but `scripts/validate-missions.sh` actually outputs `"Passed (no critical issues): N (XX%)"`. The grep never matched, making the <90% threshold check dead code — CI always exited 0 regardless of actual pass rate.

## Fix

- Added a machine-parseable `PASS_RATE=XX%` line to `validate-missions.sh` output
- Updated `.github/workflows/kb-nightly-validation.yml` to grep for `PASS_RATE=` and extract the numeric value with `grep -oP 'PASS_RATE=\K\d+'`
- Hard-fails when pass rate < 90% as originally intended

### DIFF
--- a/.github/workflows/kb-nightly-validation.yml
+++ b/.github/workflows/kb-nightly-validation.yml
@@ -28,8 +28,8 @@ jobs:
       - name: Check for regressions
         run: |
           # Fail if mission pass rate drops below 90%
-          if grep -q "pass rate" validation-results.txt; then
-            RATE=$(grep -oP '\d+%' validation-results.txt | head -1 | tr -d '%')
+          if grep -q "PASS_RATE=" validation-results.txt; then
+            RATE=$(grep -oP 'PASS_RATE=\K\d+' validation-results.txt | head -1)
             if [ "${RATE:-100}" -lt 90 ]; then
               echo "::error::Mission validation pass rate dropped below 90% (currently ${RATE}%)"
               exit 1
--- a/scripts/validate-missions.sh
+++ b/scripts/validate-missions.sh
@@ -348,8 +348,10 @@ echo -e "  Total missions validated:  ${BOLD}${total}${RESET}"
 if [[ $total -gt 0 ]]; then
   pass_pct=$((passed * 100 / total))
   echo -e "  Passed (no critical issues): ${GREEN}${passed}${RESET} (${pass_pct}%)"
+  echo "PASS_RATE=${pass_pct}%"
 else
   echo -e "  Passed: ${GREEN}0${RESET}"
+  echo "PASS_RATE=0%"
 fi
 
 echo -e "  Critical issues:           ${RED}${critical}${RESET}"


clubanderson:
**Reviewer note (post-merge):** Copilot flagged a latent issue in `scripts/validate-missions.sh`:

`PASS_RATE=...` is only emitted in the summary section (lines 351/354), but there are earlier `exit 0` paths (lines 47, 304, 383) — e.g. when no missions are found or rate-limited — that bypass the summary entirely. When the script exits early, `validation-results.txt` will not contain `PASS_RATE=`, causing any grep-based CI check on that value to silently fail or produce no match.

**Recommended fix:** emit `PASS_RATE=0%` (or `PASS_RATE=N/A`) before each early `exit 0`, or move the PASS_RATE echo to a trap/cleanup function that always runs on exit.

PR #11282, opened 2026-05-01
🐛 fix(Missions): i18n namespace separators and defer-in-loop

clubanderson:
Fixes issues introduced by PR#11242 that landed on main without post-review fixes.

**Changes:**
- `Missions.tsx`: Fix 13 `t()` calls using dot notation (`cards.X`, `common.X`) to colon namespace syntax (`cards:X`, `common:X`) per i18next convention. Affected: `missionStatus`, `clusterStatus`, `sortBy`, `dependencyAction`, `searchMissions`. Without this, all these strings silently fall back to their hardcoded defaults.
- `missions.go fetchWithCache`: Replace `defer resp.Body.Close()` inside retry loop with explicit close after `io.ReadAll` — defers accumulate until the function returns, keeping multiple HTTP response bodies open across retry attempts.

Addresses Copilot review comments on PR#11242.

### DIFF
--- a/pkg/api/handlers/missions.go
+++ b/pkg/api/handlers/missions.go
@@ -525,10 +525,10 @@ func (h *MissionsHandler) fetchWithCache(c *fiber.Ctx, cacheKey, url, logContext
 		if err != nil {
 			continue
 		}
-		defer resp.Body.Close()
 
 		limitedBody := io.LimitReader(resp.Body, missionsMaxBodyBytes)
 		body, err = io.ReadAll(limitedBody)
+		resp.Body.Close()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


clubanderson:
**Follow-up PR#11284:** The build broke after this merge with TS2345 on `Missions.tsx:429` — `t('common:searchMissions')` missing required `defaultValue` for i18next strict types. PR#11284 fixes this with `t('common:searchMissions', 'Search missions...')`.

Additionally, Copilot flagged (low priority): consider adding a unit test for `fetchWithCache`'s retry loop asserting that response bodies are closed on each attempt — e.g. via a custom `http.RoundTripper` whose `Body.Close()` increments a counter.

PR #11945, opened 2026-05-04
fix(hooks,a11y): align isDemoFallback with effectiveIsLoading; enable button-name axe rule

### LINKED ISSUE #11933: fix(a11y): accessibility violations capped at 10 rather than fixed heading order and missing ARIA labels
## Summary

`web/e2e/a11y.spec.ts` tracks accessibility violations with a threshold instead of fixing them. The test allows up to 10 heading-order WCAG violations and permanently disables four axe rules. This means screen reader users are living with known broken navigation, and the violations can grow as long as the count stays under the cap.

## Affected file

**File:** `web/e2e/a11y.spec.ts`  
**Introducing commit:** `486d84ea0`

### Heading violation cap

```typescript
// line 230-234
const MAX_KNOWN_HEADING_VIOLATIONS = 10
expect(results.violations.length, `Heading order violations exceeded threshold...`).toBeLessThanOrEqual(MAX_KNOWN_HEADING_VIOLATIONS)
```

Cards use `h3`/`h4` without an `h2` parent, breaking screen reader heading navigation.

### Permanently disabled axe rules

```typescript
rules: {
  'button-name': { enabled: false },       // buttons without accessible labels
  'select-name': { enabled: false },       // selects without accessible labels
  'aria-required-children': { enabled: false },  // structural ARIA misuse
  'aria-required-parent': { enabled: false },    // structural ARIA misuse
}
```

## Production impact

- Screen reader users cannot navigate the dashboard by heading structure.
- Some buttons and selects have no accessible name — keyboard-only and assistive technology users cannot identify or activate them.
- The cap pattern prevents regression but does not reduce existing violations. As new cards are added the count trends upward toward the cap.

## Proposed fix

1. Fix existing heading-order violations: ensure each card section has a proper `h2` before using `h3`/`h4`, or use `role="heading"` with correct `aria-level`.
2. Remove disabled axe rules one by one as the underlying violations are resolved:
   - Add `aria-label` or visible text to unlabelled buttons.
   - Add `<label>` or `aria-label` to unlabelled selects.
   - Fix structural ARIA issues flagged by `aria-required-children` / `aria-required-parent`.
3. Once violations are zero, replace the `toBeLessThanOrEqual(MAX_KNOWN_HEADING_VIOLATIONS)` assertion with `toHaveLength(0)` to prevent regressions.

oksaumya:
## Summary

- **BUG-006** (useNightlyE2EData.ts): isDemoFallback was evaluated against cacheResult.isLoading while isLoading returned to consumers used the derived effectiveIsLoading. This caused the demo badge to flash incorrectly during background refreshes. Introducing commit: 84903cc27. Fix: derive effectiveIsLoading first and use it in both fields.
- **BUG-008** (a11y.spec.ts, partial): Remove button-name from the global disableRules — the dedicated sub-test already asserts zero button-name violations. Add TODO(#11933) to the remaining 4 disabled rules. Introducing commit: a624eee7a.

## Test plan
- Nightly E2E card: demo badge should not flash during background refresh when hasCachedInitial=true
- npx playwright test e2e/a11y.spec.ts — WCAG audit now includes button-name checks
- No other hooks affected — change isolated to useNightlyE2EData.ts

Closes #11933 (partial)

Generated with Claude Code

### DIFF
--- a/web/e2e/a11y.spec.ts
+++ b/web/e2e/a11y.spec.ts
@@ -39,12 +39,11 @@ test.describe('Accessibility Audits', () => {
         const results = await new AxeBuilder({ page })
           .withTags(['wcag2a', 'wcag2aa', 'wcag21aa'])
           .disableRules([
-            'color-contrast',
-            'button-name',
-            'select-name',
-            'nested-interactive',
-            'aria-required-children',
-            'aria-required-parent',
+            'color-contrast',    // Theming system makes contrast checks unreliable in demo mode
+            'select-name',       // TODO(#11933): fix select labels, then remove
+            'nested-interactive', // TODO(#11933): fix nested interactive elements, then remove
+            'aria-required-children', // TODO(#11933): fix ARIA structure, then remove
+            'aria-required-parent',   // TODO(#11933): fix ARIA structure, then remove
           ])
           .exclude('[data-testid="chart"]') // Charts may have known issues
           .exclude('.recharts-wrapper') // Chart library exclusion
--- a/web/e2e/mission-control-dry-run.spec.ts
+++ b/web/e2e/mission-control-dry-run.spec.ts
@@ -383,10 +383,9 @@ test.describe('Mission Control Dry-Run Tests', () => {
 
       // Wait for dry-run to complete — look for completion indicators in the UI
       await expect(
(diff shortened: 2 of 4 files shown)


oksaumya:
Closing — tracking as issues instead.

PR #12632, opened 2026-05-08
🐛 Fix local cluster start button — enable kc-agent cluster detection

KPRoche:
## Summary

Fixes the local cluster start/stop buttons not appearing when using `./start-dev.sh` or `./startup-oauth.sh`. The issue was that the frontend couldn't detect local clusters because kc-agent authentication was failing.

## Root Cause

The authentication flow was broken:
1. Frontend tried to detect clusters via kc-agent's `/local-clusters` endpoint
2. kc-agent required Bearer token authentication
3. Frontend tried to fetch token from `/api/agent/token`
4. Backend required JWT auth (user not logged in yet), returned 401
5. Frontend had no token, kc-agent rejected the request
6. Cluster detection failed, start button didn't appear

## Solution

This PR implements a three-part fix:

**1. UI Logic Fix** (commit b8faffb49)
- Enable the start button for detected local clusters even when they show as unreachable
- Only disable for cloud clusters (EKS, GKE, AKS) where starting is not possible

**2. Startup Script Enhancement** (commit 527b3e5bc)
- When startup scripts auto-generate `KC_AGENT_TOKEN`, also export `KC_AGENT_TOKEN_GENERATED=true`
- This signals kc-agent to keep the origin-based auth bypass enabled for browser clients
- Allows frontend to connect before OAuth login completes (requires PR #12588)

**3. Backend Auth Fix** (commit de9b8fc96)
- Move `/api/agent/token` endpoint registration BEFORE the `/api` group so it bypasses JWT auth middleware
- Frontend can now fetch the token without requiring authentication
- Token is intended for kc-agent, which provides its own bearer validation

## How It Works

1. Startup scripts auto-generate token and export the generated flag
2. kc-agent accepts browser clients via trusted-origin auth bypass
3. Frontend fetches `/api/agent/token` (no JWT required)
4. Frontend authenticates to kc-agent and detects local clusters
5. Start/stop buttons appear for detected clusters

## Testing

- ✅ Tested locally with `./start-dev.sh`
- ✅ Backend `/api/agent/token` endpoint returns token without auth
- ✅ Frontend can fetch token and authenticate to kc-agent
- ✅ kc-agent `/local-clusters` endpoint returns detected clusters

Fixes #12444

Co-authored-by: Claude Haiku 4.5 <noreply@anthropic.com>


KPRoche:
PR superseded by #12588 and #12595 which were merged earlier today with equivalent fixes.

Both issues this PR addressed have already been fixed upstream:
- **PR #12588**: Auth bypass when token is startup-generated (kc-agent now checks KC_AGENT_TOKEN_GENERATED env var)
- **PR #12595**: Added /status endpoint to kc-agent for health probing

Our approach was to reuse existing /providers/health endpoint instead of adding /status, but the upstream solution works well. Closing to avoid duplicate merges.

PR #12713, opened 2026-05-08
🐛 Fix profile dropdown reload on first click during startup

### LINKED ISSUE #12712: 🐛 Fix profile dropdown reload on first click during startup
## Problem

When clicking the demo-user profile dropdown for the first time after the app starts (`./start-dev.sh`), the page appears to reload / the dropdown immediately closes.

## Root Cause

Two bugs in `UserProfileDropdown.tsx`:

1. **`mousedown` race with React toggle** — The outside-click handler used `mousedown`, which fires *before* React's `onClick` commits the `isOpen = true` state. During startup, background auth/demo state changes (`refreshUser()` → `setDemoMode()`) trigger React re-renders at unpredictable moments. If a re-render was in flight when `mousedown` fired, the handler could run against stale DOM state and close the dropdown immediately after it opened.

2. **All buttons missing `type="button"`** — All 10 `<button>` elements in the component had no explicit `type` attribute, which defaults to `type="submit"` in HTML. Any ancestor `<form>` (current or future) would be submitted on click, causing a full page reload.

## Fix

- Switch outside-click handler from `mousedown` → `click` (fires after React commits state)
- Gate the handler behind `if (!isOpen) return` so the listener is only attached when the dropdown is open
- Add `type="button"` to all 10 buttons in the component

## Files Changed

- `web/src/components/layout/UserProfileDropdown.tsx`

oksaumya:
## Summary

- Switch `mousedown` → `click` for the outside-click handler to prevent a race with React's concurrent re-renders during startup
- Gate the listener behind `isOpen` so it only attaches when the dropdown is open
- Add `type="button"` to all 10 buttons (prevents accidental form submission / page reload)

Fixes #12712

## Test plan

- [ ] Start app with `./start-dev.sh` (no OAuth)
- [ ] Click the demo-user profile dropdown immediately after the page loads — dropdown should open and stay open
- [ ] Click outside the dropdown — it should close
- [ ] Click the trigger button again — it should toggle closed
- [ ] Press Escape — it should close
- [ ] Verify all buttons inside the dropdown work correctly

)

### DIFF
--- a/docs/qa/qa-round2-high-medium.md
+++ b/docs/qa/qa-round2-high-medium.md
@@ -1,0 +1,268 @@
+# QA Inspection Report — Round 2 (High & Medium Issues)
+
+**Date:** 2026-05-05  
+**Scope:** Backend handlers, Netlify functions, E2E tests, CI workflows  
+**Critical issues:** Tracked separately as GitHub issues #12043–#12046
+
+---
+
+## HIGH SEVERITY
+
+### H-001 — Raw error details leaked to client in custom_resources.go
+
+**File:** `pkg/api/handlers/custom_resources.go`  
+**Lines:** 98, 102  
+**Introducing commit:** `071711f1b` (Split feedback.go and gitops.go into domain-specific files #10420)
+
+**Code:**
+```go
+// Line 98
+"error": fmt.Sprintf("forbidden: %v", err),
+
+// Line 102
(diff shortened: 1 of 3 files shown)


oksaumya:
Closing to fix DCO (missing Signed-off-by) and Hold Issue Guard check. Will reopen with fixed commits.

Issue #13232, opened 2026-05-12
test: missing Vitest coverage for useQASMFiles hook — fetch lifecycle, auth guards, and refetch behaviour


Ram04102007:
### Feature Description

The useQASMFiles hook fetches quantum QASM files with 
3 early-exit guards (isAuthenticated, isQuantumForcedToDemo, 
enabled) but has zero unit test coverage. A regression 
in any guard would silently expose quantum file listings 
to unauthenticated users or demo environments.

### Proposed Solution

Add Vitest unit tests at:
web/src/hooks/__tests__/useQASMFiles.test.ts

Covering 8 scenarios:
1. Initial loading state
2. Array response shape
3. Object response shape { files: [] }
4. Non-ok response error handling (403)
5. isAuthenticated=false — no fetch
6. isQuantumForcedToDemo=true — no fetch
7. enabled=false — no fetch
8. refetch triggers second fetch

### Feature Area

Dashboard / Cards

### Want to contribute?

- [x] I would like to work on this issue.

### Additional Context

All 8 tests written and passing locally.
All imports verified on disk before writing.
Related to LFX Mentorship #4189 test coverage scope.

Ram04102007:
This is a test coverage feature request, not a runtime bug.
No build SHA applies.

Console version for reference:
v0.3.20-nightly.20260410 (commit: 1ce7f6c)
from console.kubestellar.io

Ram04102007:
This is a test coverage feature request, not a runtime bug.
No build SHA applies.

Console version for reference:
v0.3.20-nightly.20260410 (commit: https://github.com/kubestellar/console/commit/1ce7f6c06b262da14c0b2f6825359449f2d75d6f)
from console.kubestellar.io

clubanderson:
@Ram04102007 if you do not use "fixes" with the issue number of that the PR fixes, in the PR description, then the issues will not close when your PR is merged.

PR #13233, opened 2026-05-12
test: add Vitest coverage for useQASMFiles hook — fetch lifecycle, auth guards, refetch

Ram04102007:
Related to #13232

Adds Vitest unit tests for useQASMFiles hook covering 
8 scenarios:

1. Initial loading state (isLoading=true)
2. Array response shape from API
3. Object response shape { files: [] }
4. Non-ok response error handling (403)
5. isAuthenticated=false — no fetch
6. isQuantumForcedToDemo=true — no fetch
7. enabled=false — no fetch
8. refetch triggers second fetch

Test location: web/src/hooks/__tests__/useQASMFiles.test.ts

All imports verified on disk. All 8 tests passing locally.


Ram04102007:
Fixed both issues:
- Added isQuantumForcedToDemo to useEffect 
  dependency array (line 59)
- Added dynamic demo mode transition test 
  (true → false after mount)

All 9 tests passing locally.

Issue #13832, opened 2026-05-15
📝 test: missing Vitest/RTL coverage for useQuantum* hook glue and Quantum card UI components


ANAMASGARD:
### Feature Description

The `useCachedQuantum.ts` exports four public hooks and the `web/src/components/cards/quantum/`
folder contains three UI components — none of which have any test coverage on main today:

**Hooks with no glue-level tests:**
- useQuantumSystemStatus
- useQuantumAuthStatus
- useQuantumCircuitAscii
- useQuantumQubitGridData

**UI components with no test files at all:**
- QuantumStatus.tsx
- QuantumCircuitViewer.tsx
- QuantumQubitGrid.tsx

### Proposed Solution

Would cover:

Hook tests (Vitest + renderHook):
1. useQuantumSystemStatus — unauthenticated path returns disabled result (data: null, all flags false, refetch wired)
2. useQuantumSystemStatus — isDemoData = true only when isDemoFallback=true AND isLoading=false
3. useQuantumAuthStatus — unauthenticated path returns DEFAULT_AUTH_STATUS with all flags false
4. useQuantumAuthStatus — demo mode returns DEFAULT_AUTH_STATUS and isDemoData follows isDemoFallback && !isLoading
5. useQuantumCircuitAscii — unauthenticated path returns disabled result with data: null
6. useQuantumCircuitAscii — demo mode returns DEMO_QUANTUM_CIRCUIT and isDemoData = true
7. useQuantumQubitGridData — unauthenticated path returns disabled result with data: null
8. useQuantumQubitGridData — demo mode maps DEMO_QUANTUM_QUBITS and versionInfo correctly from DEMO_QUANTUM_STATUS

UI component tests (Vitest + RTL):
9. QuantumStatus — renders skeleton when isLoading=true
10. QuantumStatus — renders error state when isFailed=true
11. QuantumStatus — renders demo badge when isDemoData=true
12. QuantumStatus — renders version info text when data.version_info is populated
13. QuantumCircuitViewer — renders ASCII circuit inside monospace block when circuitAscii is provided
14. QuantumCircuitViewer — renders empty state message when circuitAscii is null or empty
15. QuantumQubitGrid — renders correct qubit count and pattern when qubits data is present
16. QuantumQubitGrid — renders meaningful "no qubits" fallback when qubits is null

### Feature Area

Dashboard / Cards

### Want to contribute?

- [x] I would like to work on this issue.

ANAMASGARD:
[...]

PR #14012, opened 2026-05-16
test(quantum): RTL coverage for QuantumCircuitViewer and QuantumQubitGrid components (#13832)

ANAMASGARD:
Refs #13832

Adds Vitest + RTL coverage for the quantum circuit and qubit grid card UIs (tests 13–16 from the issue).

**QuantumCircuitViewer.test.tsx**
- Renders ASCII circuit inside `pre.quantum-circuit-display` when `circuitAscii` is provided.
- Renders fallback `Unable to load quantum circuit diagram` when hook `data` is null.

**QuantumQubitGrid.test.tsx**
- Renders qubit count, pattern, and backend version inside the blue info panel when nested `data.qubits` / `data.versionInfo` are present (queries scoped with `within` + `exact: true` to avoid mask `<option>` label collisions).
- Renders logo-mode fallback (`num_qubits` 8, no `01010` pattern) and an `svg` when hook `data` is null.

Build and lint are validated by CI per project guidelines.

### DIFF
--- a/web/src/components/cards/quantum/__tests__/QuantumCircuitViewer.test.tsx
+++ b/web/src/components/cards/quantum/__tests__/QuantumCircuitViewer.test.tsx
@@ -1,0 +1,70 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { render, screen } from '@testing-library/react'
+
+const mockUseQuantumCircuitAscii = vi.fn()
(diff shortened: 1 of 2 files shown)


clubanderson (approved):
RTL coverage for QuantumCircuitViewer and QuantumQubitGrid looks good — ASCII rendering, null/empty fallbacks, qubit grid info box scoping all well tested.

Issue #13913, opened 2026-05-15
bug: updateSingleClusterInCache skips metric sharing when all updates are zero due to falsy || check in shared.ts


Ram04102007:
**Type:** bug  
**Target:** Console Application  
**Component:** `web/src/hooks/mcp/shared.ts` — Line 594

The bug exists in the current main branch and can be verified at:
https://github.com/kubestellar/console/blob/main/web/src/hooks/mcp/shared.ts#L594

## Describe the Bug

`updateSingleClusterInCache` uses `||` to check whether any metric
fields were updated before calling `shareMetricsBetweenSameServerClusters`:

```ts
// shared.ts:594 — UNSAFE
if (updates.nodeCount || updates.podCount || updates.cpuCores || updates.memoryGB ||
    updates.storageGB || updates.cpuRequestsCores || updates.memoryRequestsGB) {
  updatedClusters = shareMetricsBetweenSameServerClusters(updatedClusters)
}
```

All fields are numeric. When a cluster scales to zero, all values are `0`.
`0 || 0 || undefined || ...` evaluates to `false`, so
`shareMetricsBetweenSameServerClusters` is never called.

Alias clusters (e.g. `prow` and its full OpenShift context name pointing
to the same server) are then not updated and retain stale non-zero
node/pod counts from the previous sample — making a scaled-to-zero
cluster appear still running in the dashboard.

## Concrete Failure Case

A cluster with two kubeconfig contexts (`prow` and
`default/api-prow.openshiftapps.com:6443/...`) scales down to 0 nodes.

`updateSingleClusterInCache('prow', { nodeCount: 0, podCount: 0 })` is called.

The condition evaluates as `0 || 0` → `false`.
`shareMetricsBetweenSameServerClusters` is skipped.
The alias retains `nodeCount: 10` from the previous sample.

| updates | Condition | shareMetrics called | Alias updated | Correct? |
|---|---|---|---|---|
| `{ nodeCount: 5 }` | `5` → truthy | ✅ yes | ✅ yes | ✓ |
| `{ nodeCount: 0, podCount: 0 }` | `0 \|\| 0` → false | ❌ no | ❌ no | ✗ |
| `{ nodeCount: 0 }` | `0` → false | ❌ no | ❌ no | ✗ |

## Why This Is Confirmed Real — Not Theoretical

The developers already knew `0` is a valid metric value and fixed it
elsewhere in the same file. Lines 283–289 define `pickMetric` with
an explicit comment:

```ts
[...]

PR #14134, opened 2026-05-16
fix: use === undefined instead of falsy check in dedup.ts shareMetricsBetweenSameServerClusters

### LINKED ISSUE #14132: bug: dedup.ts shareMetricsBetweenSameServerClusters overwrites explicit zero podCount/nodeCount with stale alias values
**Type:** bug  
**Component:** `web/src/hooks/mcp/dedup.ts` — Lines 52–53

The bug exists in the current main branch and can be verified at:
https://github.com/kubestellar/console/blob/main/web/src/hooks/mcp/dedup.ts#L52

SHA: 3f1f99745

## Describe the Bug

`dedup.ts` contains its own copy of `shareMetricsBetweenSameServerClusters`
with the same falsy trap pattern as the one fixed in `shared.ts` (issue #13913):

```ts
// dedup.ts:52–53 — UNSAFE
if (!cluster.nodeCount || cluster.nodeCount === 0)
if (!cluster.podCount  || cluster.podCount  === 0)
```

Both conditions treat an explicit `0` as "no data available — copy from alias."
When a cluster scales to zero nodes or pods, its `nodeCount: 0` and
`podCount: 0` are valid domain values — but this guard overwrites them
with stale non-zero counts from the alias cluster.

## Concrete Failure Case

A cluster scales down to 0 pods. The dedup pipeline runs.
`cluster.podCount` is `0` — a valid explicit value.

`!cluster.podCount` evaluates to `true` (falsy trap).
The condition fires and copies the alias's stale `podCount: 10`
back over the fresh `0`.

The dashboard shows the scaled-to-zero cluster as still running
with 10 pods — silent wrong data, no error surfaced.

## Why This Is Confirmed Real — Not Theoretical

The existing test at `dedup-coverage.test.ts:35` was encoding
the **buggy behaviour** — it asserted that an explicit `podCount: 0`
gets overwritten by the alias value, treating this incorrect
overwrite as expected behaviour.

This is the same pattern as issue #13913 (`shared.ts:594`)
which was confirmed and fixed by kubestellar-hive.
`dedup.ts` is a parallel implementation with the identical flaw.

## Environment

- **File:** `web/src/hooks/mcp/dedup.ts` lines 52–53
- **Trigger:** Any cluster that scales to 0 nodes or pods
  while sharing a server with an alias cluster in the dedup pipeline

## Steps To Reproduce

1. Connect two kubeconfig contexts pointing to the same server
2. Scale the cluster to 0 pods
3. Run the dedup pipeline
4. Observe: alias cluster copies stale pod count back over the zero
5. Dashboard shows cluster as still running — silent wrong data

## Expected Behavior

An explicit `podCount: 0` or `nodeCount: 0` is a valid domain value.
It must be preserved. Alias values should only fill in when the
field is `undefined` — not when it is explicitly zero.

## Browser Console / Backend Logs

No errors — stale non-zero values silently overwrite the correct zero.
Dashboard shows cluster as running with no error surfaced.

## Additional Context

Fix — change falsy checks to explicit `=== undefined`:

```ts
// BEFORE
if (!cluster.nodeCount || cluster.nodeCount === 0)
if (!cluster.podCount  || cluster.podCount  === 0)

// AFTER
if (cluster.nodeCount === undefined)
if (cluster.podCount  === undefined)
```

Also corrects the existing test at `dedup-coverage.test.ts:35`
which was asserting the buggy overwrite behaviour as expected —
renamed and fixed to assert that explicit `podCount: 0` is preserved.

This is the same fix pattern applied to `shared.ts:594` in issue #13913,
now applied to the parallel implementation in `dedup.ts`.

Ram04102007:
Fixes #14132

## What Changed

Changed falsy `||` checks to explicit `=== undefined` in
`shareMetricsBetweenSameServerClusters` in `dedup.ts`:

```ts
// BEFORE
if (!cluster.nodeCount || cluster.nodeCount === 0)
if (!cluster.podCount  || cluster.podCount  === 0)

// AFTER
if (cluster.nodeCount === undefined)
if (cluster.podCount  === undefined)
```

Also corrected `dedup-coverage.test.ts:35` which was asserting
the buggy overwrite behaviour as expected — renamed and fixed
to assert that explicit `podCount: 0` is preserved.

## Why This Fix Is Correct

Same falsy trap pattern as `shared.ts:594` fixed in issue #13913.
`dedup.ts` contains a parallel implementation of
`shareMetricsBetweenSameServerClusters` with the identical flaw.

When a cluster scales to 0 pods, `!cluster.podCount` evaluates
to `true`, causing stale alias values to overwrite the valid zero.

## Test Results

128 tests passing, 0 failures

### DIFF
--- a/web/src/hooks/mcp/__tests__/dedup-coverage.test.ts
+++ b/web/src/hooks/mcp/__tests__/dedup-coverage.test.ts
@@ -32,7 +32,7 @@ describe('dedup.ts extended coverage', () => {
       expect(shareMetricsBetweenSameServerClusters(undefined as unknown as ClusterInfo[])).toEqual([])
     })
 
-    it('copies pod count from source when target has zero', () => {
+    it('does not overwrite explicit podCount:0 with stale source data', () => {
       const withPods = makeCluster({
         name: 'full',
         server: 'https://api.example.com',
@@ -47,7 +47,8 @@ describe('dedup.ts extended coverage', () => {
       })
       const result = shareMetricsBetweenSameServerClusters([noPods, withPods])
       const alias = result.find(c => c.name === 'alias')!
-      expect(alias.podCount).toBe(50)
+      // podCount:0 is an explicit value (scaled-to-zero), not missing — must be preserved
+      expect(alias.podCount).toBe(0)
     })
 
     it('copies memory and storage metrics via nullish coalescing', () => {
--- a/web/src/hooks/mcp/dedup.ts
+++ b/web/src/hooks/mcp/dedup.ts
@@ -49,8 +49,8 @@ export function shareMetricsBetweenSameServerClusters(clusters: ClusterInfo[] | 
     if (!source) return cluster
 
     // Check if we need to copy anything - include nodeCount, podCount, and capacity/requests
-    const needsNodes = (!cluster.nodeCount || cluster.nodeCount === 0) && source.nodeCount && source.nodeCount > 0
-    const needsPods = (!cluster.podCount || cluster.podCount === 0) && source.podCount && source.podCount > 0
+    const needsNodes = cluster.nodeCount === undefined && source.nodeCount !== undefined && source.nodeCount > 0
+    const needsPods = cluster.podCount === undefined && source.podCount !== undefined && source.podCount > 0
     const needsCapacity = !cluster.cpuCores && source.cpuCores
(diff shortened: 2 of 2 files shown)


Ram04102007:
This applies the same fix pattern as issue #13913 (shared.ts:594) to the parallel implementation in dedup.ts. All 128 tests pass and no conflicts. Could a maintainer please review?

Issue #15946, opened 2026-05-28
Quantum Control Panel: softer auth-error UX + suppress IBM auth check on local backends


KPRoche:
## Problem

The Quantum Control Panel polls `/api/quantum/auth/status` every 8 seconds regardless of which backend is selected. When IBM Quantum's upstream API has a transient hiccup (rate limit, instance-list 5xx, timeout, "max retries attempted"), our proxy forwards the non-2xx response verbatim, the auth-status fetcher throws the raw response body as the `Error.message`, and the Control Panel renders that JSON blob in a red "Refresh failed" banner — alarming users into deleting/re-entering tokens that are actually fine.

Worse, the check fires even when the user has selected the local **Aer Simulator** or **QASM Simulator**, neither of which need IBM at all. The user sees scary auth errors during purely-local work.

A second usability gap: the credentials badge today is binary (`Configured` vs `Not configured`) and is keyed on the most recent live auth check. After a pod restart, the persistent volume retains the saved token but the in-memory validation flag is fresh, so the badge falsely flips to `Not configured` until the next successful validation — making users think their token is gone.

## Proposed fix

1. **Suppress** the auto-poll auth-status check when the selected backend doesn't need IBM (only `qx5`, `least`, `aer_noise` do).
2. **Classify** auth-status errors so transient upstream issues render as a soft yellow "IBM Quantum upstream unavailable, retrying…" banner instead of a red error.
3. **Three-state badge**:
   - `🛡️ Configured` (green) — validation succeeded in this browser session.
   - `✓ Stored` (blue) — token likely exists on the backend but not validated this session (covers post-restart, non-IBM backend, transient blip).
   - `Not configured` (gray) — no token saved.
4. **Validate-on-save** unconditionally — refetch auth status after a token save, regardless of selected backend.
5. **Default to Stored after pod restart** — the Configured state requires a session-local successful observation, not just persisted cache state.
6. **Validate now button** next to the Stored badge to manually trigger a one-shot check.

## Acceptance

- No red banner for transient IBM upstream errors (rate-limit / 5xx / timeout / max-retries).
- No auth-status polling when on `aer` / `sim`.
- Badge correctly distinguishes "we know it works" from "a token is saved but unvalidated."
- IBM-only backend options (`least`, `aer_noise`) appear in the dropdown when state is `stored` or `configured`, not just `configured`.

PR #15960, opened 2026-05-28
✨ Consume tokenStored + lastIbmError from quantum-kc-demo v0.4.0

KPRoche:
## Summary

Replaces the client-side `tokenLikelyStored` inference with the explicit `tokenStored` field that quantum-kc-demo `v0.4.0` now returns on `/api/quantum/auth/status`. Also threads through the new structured `lastIbmError` payload so the workload's authoritative classification drives the yellow transient banner instead of message-text matching.

**Closes Bug 1** from `kubestellar-hive[bot]`'s post-merge review of #15948 — *"the new Stored state can be shown even when no credentials exist."* That bug was a symptom of inferring storage state from the cache's `lastSuccessfulRefresh`: a fetch returning `{authenticated:false}` after a `clear` is still a successful fetch, so `lastRefresh` updated and the badge fell back to "Stored". With v0.4.0 the workload reports `tokenStored` explicitly, and after `clear` it's `false` regardless of cache state.

The workload-side change shipped in `KPRoche/quantum-kc-demo` v0.4.0 (image `ghcr.io/kproche/quantum-kc-demo:v0.4.0`, deployed and verified live on the kind-quantum cluster). This PR consumes those new fields on the Console side.

## What changed

**`useCachedQuantum.ts`:**
- Adds `tokenStored: boolean` and `lastIbmError: QuantumIbmError | null` to the `QuantumAuthStatus` interface.
- `fetchQuantumAuthStatus()` coerces missing `tokenStored` to `false` so pre-v0.4 workloads still work — badge sits at "Not configured" until next successful validation, harmless and self-healing.

**`QuantumControlPanel.tsx`:**
- Replaces `tokenLikelyStored = lastAuthRefresh !== null || ibmAuthenticated` with the explicit `ibmTokenStored = authStatus.tokenStored`.
- Drops the now-unused `lastAuthRefresh` destructure and `hasHistoricalValidation` derived value.
- Prefers `authStatus.lastIbmError.retryable` over `classifyApiError(authStatusError)` when classifying for the yellow banner. Keeps `classifyApiError` as a fallback for older workloads and pre-body network errors.

**Backward compatibility:** older workloads (pre-v0.4) that omit `tokenStored` and `lastIbmError` are handled — `tokenStored` coerces to `false`, the existing `classifyApiError` fallback covers banner classification. The badge sits at "Not configured" against an old workload until the first successful validation.

## Tests

- 6 quantum test files, **50/50 passing locally** (`vitest run src/components/cards/quantum src/hooks/__tests__/useQuantumAuthStatus.test.ts src/hooks/__tests__/useCachedQuantum-pure.test.ts`).
- New test cases:
  - Bug 1 regression: `tokenStored:false` + recent `lastRefresh` must yield "Not configured", not "Stored".
  - Pre-v0.4 compat: workload omitting `tokenStored` is treated as "Not configured".
  - Yellow banner driven by `lastIbmError.retryable:true` (v0.4.0+ workload, no error string).
  - Red banner driven by `lastIbmError.retryable:false`.
  - Updated fetcher pure tests for the new return shape, plus a v0.4.0+ passthrough case and a coercion case.

## Addresses Copilot nits from #15957

1. **Explicit backend selection.** The non-IBM red-banner test now calls `selectNonIbmBackend(container)` explicitly rather than relying on the component's default backend remaining `aer`. Mirrors the existing `selectIbmBackend()` helper used for IBM-backend tests.
2. **Stable selectors.** Banner assertions now target `data-testid="quantum-control-panel-fatal-banner"` and `quantum-control-panel-transient-banner` instead of the raw error text or i18n key. Future banner-content refactors won't silently invalidate the tests. Added matching `role="alert"` / `role="status"` attributes for accessibility.

## Test plan

- [x] `vitest run src/components/cards/quantum src/hooks/__tests__/useQuantumAuthStatus.test.ts src/hooks/__tests__/useCachedQuantum-pure.test.ts` — all green
- [ ] Visual regression baseline (`app-quantum-control-panel-demo.png`) unchanged — banner JSX only added test-ids/roles, no visual changes
- [ ] Manual: connect to quantum-kc-demo v0.4.0 cluster, verify badge transitions: clear → "Not configured", save token → "Configured", restart pod → "Stored", validate-now → "Configured"

Refs #15946, #15948, #15957

### DIFF
--- a/web/src/components/cards/quantum/QuantumControlPanel.tsx
+++ b/web/src/components/cards/quantum/QuantumControlPanel.tsx
@@ -112,7 +112,6 @@ export const QuantumControlPanel: React.FC = () => {
     data: authStatus,
     isRefreshing: isAuthRefreshing,
     error: authStatusError,
-    lastRefresh: lastAuthRefresh,
     refetch: refetchAuthStatus,
   } = useQuantumAuthStatus({
     isAuthenticated,
@@ -122,6 +121,8 @@ export const QuantumControlPanel: React.FC = () => {
   })
 
   const ibmAuthenticated = authStatus.authenticated
+  const ibmTokenStored = authStatus.tokenStored
+  const lastIbmError = authStatus.lastIbmError
 
   // Mark the session as validated whenever a successful auth check returns
   // authenticated:true. This is what flips the badge from "Stored" → "Configured".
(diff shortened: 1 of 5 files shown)


KPRoche:
Addressed both Copilot review findings in `3a0557f29`:

**Finding 1 — `lastIbmError` runtime validation** (`useCachedQuantum.ts:198`, `212`)

Added `coerceLastIbmError(raw: unknown): QuantumIbmError | null` at the fetcher boundary. Required fields: `code` (string), `message` (string), `retryable` (boolean). Anything missing or wrong-typed coerces to `null`, which keeps the `classifyApiError` message-text fallback alive instead of letting a partial payload silently suppress it.

Unrecognized `code` values are intentionally accepted as-is — only the *structural* shape is validated, not the enum — so a future workload that adds a new error code doesn't silently lose its classification on older Console builds.

11 new test cases cover this: 9 malformed-payload variants (missing fields, wrong types, non-object payloads), 1 forward-compat case for unrecognized codes, 1 explicit-`null` case.

**Finding 2 — banner content assertions weakened** (`QuantumControlPanel.test.tsx:183`, `198`)

Fair catch — when I migrated from raw-text to `data-testid` selectors (addressing the prior Copilot nit on #15957), I overcorrected and dropped content checks entirely. Those tests would have passed against an empty / wrong-content banner.

Now both: testid for stability against i18n renames, AND `toHaveTextContent` for content verification. Applied to all four flagged tests.

61/61 quantum tests pass locally.

Issue #17226, opened 2026-06-08
Stellar SSE stream panics with nil pointer dereference on /api/stellar/stream


KPRoche:
## Symptom

Backend crashes with SIGSEGV the moment a client connects to `/api/stellar/stream` (the Stellar SSE endpoint).

```
09:52:31 | 200 |     1.07376ms | GET /api/stellar/stream
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1c497f6]

goroutine 682 [running]:
github.com/valyala/fasthttp.(*RequestCtx).UserValue(...)
        .../fasthttp@v1.58.0/server.go:709
github.com/gofiber/fiber/v2.(*Ctx).Locals(...)
        .../fiber/v2@v2.52.13/ctx.go:985 +0x36
github.com/kubestellar/console/pkg/api/middleware.GetUserID(...)
        .../pkg/api/middleware/auth.go:721
github.com/kubestellar/console/pkg/api/handlers.(*StellarHandler).Stream.func1(0x699b06eb1c0)
        .../pkg/api/handlers/stellar_observations.go:399 +0x205
github.com/valyala/fasthttp.NewStreamReader.func1()
        .../fasthttp@v1.58.0/stream.go:44 +0x39
created by github.com/valyala/fasthttp.NewStreamReader in goroutine 86
        .../fasthttp@v1.58.0/stream.go:43 +0x285
```

The HTTP 200 logs because the parent handler returns successfully — the panic happens in the SSE goroutine that fasthttp spawns from `SetBodyStreamWriter`.

## Root cause

`pkg/api/handlers/stellar_observations.go:399` (introduced in #17008) calls `middleware.GetUserID(c)` from inside the `SetBodyStreamWriter` goroutine:

```go
c.Context().SetBodyStreamWriter(func(w *bufio.Writer) {
    ...
    if userStore, ok := h.store.(store.Store); ok {
        resolvedUser, resolveErr := userStore.GetUser(streamCtx, middleware.GetUserID(c))
        ...
    }
    ...
})
```

`GetUserID(c)` calls `c.Locals("userID")`, which reads from `c.Context()` (the underlying `*fasthttp.RequestCtx`). By the time the stream goroutine runs, the parent request handler has returned and fasthttp has recycled the `RequestCtx` — so its internal `userValues` map is nil and `UserValue` segfaults at `fasthttp/server.go:709`.

## Fix

[...]

PR #17233, opened 2026-06-08
[quality] add regression test for Stellar SSE stream nil-pointer panic (#17226)

clubanderson:
## Test Improvement

Adds regression test for the Stellar SSE stream nil-pointer panic (#17226, fixed by #17227).

### Tests added:
- **`TestStellarStream_NoNilPointerPanic`** — verifies the `/api/stellar/stream` endpoint produces a valid SSE response without panicking when the fasthttp `*RequestCtx` is recycled after the `SetBodyStreamWriter` goroutine starts. This was the root cause of the SIGSEGV in #17226.
- **`TestStellarStream_Unauthenticated`** — verifies 401 for missing user identity.

### Context

PR #17227 fixes the panic by capturing `userID` as a UUID in the parent handler scope before entering the stream goroutine. This test prevents the regression from being reintroduced.

**Note:** This test will fail on `main` until #17227 merges. It should be merged after or alongside that fix.

## Related Issue
Regression test for #17226

---
*Filed by quality agent (hold-gated mode). Human review required.*

### DIFF
--- a/pkg/agent/server_federation_test.go
+++ b/pkg/agent/server_federation_test.go
@@ -98,11 +98,11 @@ func (f *handlerFakeProvider) ReadPendingJoins(_ context.Context, _ *rest.Config
 	return nil, nil
 }
 
-// newTestServer returns a *Server wired with a real (empty) k8s client and
+// newFederationTestServer returns a *Server wired with a real (empty) k8s client and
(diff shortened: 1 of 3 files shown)


clubanderson:
Closing: broken by newTestServer signature change from PR #17257. Tracked in #17258. Will be re-created with correct signature.

PR #18567, opened 2026-06-17
[scanner] fix: resolve test regressions in middleware JWT and auth cleanup

### LINKED ISSUE #18562: [quality] test regressions: middleware JWT type assertion, audit SSRF, auth cleanup count
## Finding

Several test suites have regressions caused by production code evolution that left test expectations stale:

### 1. `pkg/api/middleware` — `TestParseJWT_InvalidClaims`
**Root cause:** `ParseJWT()` always calls `ParseWithClaims(token, &UserClaims{}, ...)`, so the returned `token.Claims` is always `*UserClaims` type — even when the original JWT only had `RegisteredClaims`. The test asserts `_, ok := token.Claims.(*UserClaims); assert.False(t, ok)` which will never be true.

**Fix:** Assert that UserClaims-specific fields (UserID, Email) are zero-value instead of checking the type assertion.

### 2. `pkg/api/audit` — 3 SSRF failures
Tests `TestRegisterDestinationSplunkWithFullConfig`, `TestRegisterDestinationElasticWithURL`, and `TestBuildSummary` use `127.0.0.1` URLs. The SSRF protection layer (added after these tests) correctly blocks private IPs. Tests need to either:
- Use a test-mode SSRF bypass
- Use `httptest.NewServer()` with an allowed IP
- Mock the SSRF validator

### 3. `pkg/api/handlers/auth` — `TestAuthHandler_CleanupExpiredTokens`
Expects cleanup of 2 expired tokens, but 3 are returned. Likely a test DB isolation issue (shared DB accumulates tokens from prior subtests) or production logic change adding an additional expired token.

## Recommendation

1. Fix middleware test to assert on field values, not type
2. Fix audit tests to use SSRF-safe test helpers (e.g., `ssrf.NewAllowAllValidator()` for test mode)
3. Fix auth cleanup test DB isolation or update expected count

## Priority
- Impact: medium (tests are red on CI, masking real regressions)
- Effort: low (each fix is 2-5 lines)

---

clubanderson:
Fixes #18562

## Changes

Resolves test regressions in middleware JWT and auth cleanup:

### 1. `pkg/api/middleware/auth_test.go` — `TestParseJWT_InvalidClaims`
- **Root cause:** `ParseJWT()` always calls `ParseWithClaims(token, &UserClaims{}, ...)`, so `token.Claims` is always `*UserClaims` — the type assertion always succeeds.
- **Fix:** Changed to assert that UserClaims-specific fields (`UserID`, `GitHubLogin`) are zero-value instead of checking the type assertion result.

### 2. `pkg/api/handlers/auth/handler_test.go` — `TestAuthHandler_CleanupExpiredTokens`
- **Root cause:** Test uses `time.Now()` (local time) but `CleanupExpiredTokens()` uses `time.Now().UTC()` for comparison. In non-UTC timezones, stored local timestamps can compare differently against UTC.
- **Fix:** Changed test to use `time.Now().UTC()` to match the production code's comparison.

### Note
The SSRF audit test failures (part 3 of #18562) are addressed separately in PR #18563.

### DIFF
--- a/pkg/api/handlers/auth/handler_test.go
+++ b/pkg/api/handlers/auth/handler_test.go
@@ -210,7 +210,7 @@ func TestAuthHandler_CleanupExpiredTokens(t *testing.T) {
 	jti2 := uuid.New().String()
 	jti3 := uuid.New().String()
 
-	now := time.Now()
+	now := time.Now().UTC()
 	expiredTime := now.Add(-time.Hour)
 	futureTime := now.Add(time.Hour)
 
--- a/pkg/api/middleware/auth_test.go
+++ b/pkg/api/middleware/auth_test.go
@@ -973,9 +973,14 @@ func TestParseJWT_InvalidClaims(t *testing.T) {
 	require.NoError(t, err)
 	assert.NotNil(t, token)
 
-	// The token is parsed but claims won't match UserClaims type
-	_, ok := token.Claims.(*UserClaims)
-	assert.False(t, ok, "claims should not be UserClaims type")
+	// ParseJWT always uses &UserClaims{}, so type assertion succeeds,
+	// but UserClaims-specific fields should be zero-value for a standard claims token
+	uc, ok := token.Claims.(*UserClaims)
+	assert.True(t, ok, "claims should be *UserClaims (ParseJWT always uses UserClaims)")
+	if ok {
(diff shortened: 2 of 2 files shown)


clubanderson:
Superseded by #18570 which takes a clean approach to both fixes. Closing.

PR #19167, opened 2026-06-19
🐛 fix: KB nightly validation schema fixture generation

### LINKED ISSUE #19159: [ci-maintainer] KB Nightly Validation: schema test fixture gives false-positive exit 0 — schema validation being silently skipped
## CI Issue

The **KB Nightly Validation** workflow (`kb-nightly-validation.yml`) is failing at step 6 **"Test mission schema validation path"** with a false-positive exit code. The test fixture that should trigger a schema failure (missing `version` field) instead causes the script to exit 0.

## Evidence

**Workflow**: KB Nightly Validation  
**Run ID**: 27739018171 (main `d44da30655b9104c73f1d3cc76684dee17cbff82`, 2026-06-18T05:34 UTC)  
**Job**: validate-kb → step 6 "Test mission schema validation path"

**Log output**:
```
✓ accepts minimal runtime-valid mission fixture
✗ rejects mission missing required version
  Expected schema validation to fail.
Process completed with exit code 1.
```

**Note**: This is a **different failure** from issue #19072 (which describes step 11 `gh issue create` crashing). The primary blocker in run 50 is step 6, which causes steps 7–10 to be skipped.

## Root Cause

The test script `scripts/validate-missions-schema-test.sh` calls:
```bash
./scripts/validate-missions.sh --local scripts/testdata/validate-missions/invalid-missing-version/ --schema web/src/lib/missions/mission.schema.json
```

This call is expected to EXIT 1 (schema validation failure), but it exits 0 (success). The most likely cause is that `ajv_formats_available()` inside `validate-missions.sh` is returning non-zero (its smoke test is failing), which causes `SCHEMA_FILE` to be cleared and **schema validation to be silently skipped**. When AJV schema validation is skipped, the invalid mission only goes through manual field checks (which pass since it has a `title` and `steps`), causing exit 0.

The `ajv_formats_available` smoke test runs:
```bash
ajv validate --spec=draft7 -s "$SCHEMA_FILE" -d "$smoke_file" -c ajv-formats
```

If this fails for any reason (e.g., `mission.schema.json` recently changed to require fields not present in the smoke fixture, or `-c ajv-formats` fails to load the plugin), schema validation is silently disabled.

## Recommendation

1. Add an explicit `--fail-on-schema-skip` guard to `validate-missions-schema-test.sh`:  if AJV is unavailable or skipped, the test script should exit 2 (configuration error) rather than continuing
2. Or: explicitly check that `command -v ajv` succeeds before calling `validate-missions-schema-test.sh` in the workflow
3. Investigate why `ajv_formats_available` is returning non-zero in CI since run 50 started failing (runs 41-49 succeeded)

Previous runs (41–49, June 9-17) all passed, confirming this is a regression introduced between June 17 and June 18.

---

clubanderson:
Fixes #19159

## Problem
The KB Nightly Validation workflow's schema test fixture generation was failing silently. When the AJV schema validation check (`ajv_formats_available()`) failed for any reason, the schema validation was silently skipped without warning the test script. This caused invalid missions (e.g., missing required `version` field) to pass validation during the fixture tests.

## Root Cause
The `validate-missions-schema-test.sh` script was calling `validate-missions.sh` with schema validation enabled, but when `ajv_formats_available()` smoke test failed, the `SCHEMA_FILE` variable was silently cleared in `validate-missions.sh` (lines 276-279), causing the script to return 0 even for invalid missions. The test script had no way to detect that schema validation was skipped.

## Solution
Add an explicit `check_schema_validation_available()` function to `validate-missions-schema-test.sh` that:
1. Runs the valid fixture test and captures output
2. Checks if schema validation was skipped (by detecting "schema validation skipped" in output)
3. Exits with status 2 (configuration error) if schema validation was skipped
4. Ensures schema validation is actually running before proceeding with fixture tests

This prevents false-positive passes for invalid missions and clearly communicates when dependencies are missing or misconfigured.

## Testing
- Schema validation tests now fail if AJV or ajv-formats plugin is unavailable
- Valid missions still pass (smoke test fixture includes all required fields)
- Invalid missions (missing version) are properly rejected when schema validation is available

### DIFF
--- a/scripts/validate-missions-schema-test.sh
+++ b/scripts/validate-missions-schema-test.sh
@@ -9,6 +9,17 @@ VALID_DIR="${TESTDATA_ROOT}/valid"
 INVALID_DIR="${TESTDATA_ROOT}/invalid-missing-version"
 EXPECTED_VERSION_ERROR="must have required property 'version'"
 
+check_schema_validation_available() {
+  local output
+  output=$(./scripts/validate-missions.sh --local "$VALID_DIR" --schema "$SCHEMA_FILE" 2>&1 || true)
+  
+  if [[ "$output" == *"schema validation skipped"* ]]; then
+    echo "ERROR: Schema validation was skipped. AJV tools may not be available or properly configured." >&2
+    echo "$output" >&2
+    exit 2
+  fi
+}
+
 assert_validation_passes() {
   local name="$1"
   local mission_dir="$2"
@@ -46,6 +57,7 @@ assert_validation_fails_with() {
   echo "✓ $name"
 }
 
+check_schema_validation_available
 assert_validation_passes "accepts minimal runtime-valid mission fixture" "$VALID_DIR"
 assert_validation_fails_with "rejects mission missing required version" "$INVALID_DIR" "$EXPECTED_VERSION_ERROR"
 


clubanderson:
Closing — superseded by #19214 (lint fixes) or no longer needed after build-gate unblock.

PR #19188, opened 2026-06-19
[scanner] 🐛 fix: resolve ESLint react-hooks errors in registerHooks

clubanderson:
Fixes the 2 ESLint errors in `web/src/lib/unified/registerHooks/` that are blocking the build-gate CI check on all open PRs.

**Errors fixed:**

1. **`customHooks.ts` line 44** (`react-hooks/purity`): Calling `Date.now()` during render
   - **Fix**: Moved `Date.now()` into state that updates via `useEffect` with `setInterval`, keeping the cutoff time pure during renders
   
2. **`demoSupport.ts` line 46** (`react-hooks/set-state-in-effect`): Synchronous `setState` in effect body
   - **Fix**: Wrapped `setState` calls in `queueMicrotask()` to defer execution and avoid cascading renders

**Verification:**
```bash
cd web && npx eslint src/lib/unified/registerHooks/customHooks.ts src/lib/unified/registerHooks/demoSupport.ts
# ✓ No errors
```

**Behavior preserved:**
- `useRecentEvents` still filters events from the last hour, cutoff updates every minute
- `useDemoDataHook` loading state behavior unchanged, just async instead of sync setState

### DIFF
--- a/web/src/lib/unified/registerHooks/customHooks.ts
+++ b/web/src/lib/unified/registerHooks/customHooks.ts
@@ -1,0 +1,121 @@
+/**
+ * Manually composed unified hooks.
+ */
+
+import { useMemo, useState, useEffect } from 'react'
+import { MS_PER_HOUR } from '../../constants/time'
+import { useCachedEvents } from '../../../hooks/useCachedData'
+import { useFluxStatus } from '../../../components/cards/flux_status/useFluxStatus'
+import { useContourStatus } from '../../../components/cards/contour_status/useContourStatus'
+import { useChaosMeshStatus } from '../../../components/cards/chaos_mesh_status/useChaosMeshStatus'
+import { DEMO_NAMESPACE_EVENTS } from './demoHooks'
+
+// ============================================================================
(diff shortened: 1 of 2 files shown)


clubanderson:
🤖 **Scanner observation**: This PR fixes ESLint errors that are blocking `build-gate` on ALL open PRs (circular dependency — the fix itself can't pass build-gate because build-gate requires these fixes). Needs admin merge bypass or review comment resolution to unblock the entire PR queue.

clubanderson:
🤖 **Scanner root-cause analysis**: The `build-gate` check is being **cancelled due to a 10-minute timeout** in `.github/workflows/pre-merge-gate.yml`. The job sequence (Go build ~1.5m + Frontend build ~5.7m + Frontend lint ~3m) exceeds the limit.

**Fix needed**: Increase `timeout-minutes` from `10` to `15` in `pre-merge-gate.yml`. This requires the `workflow` scope which the scanner bot doesn't have.

This timeout issue is blocking ALL open PRs from passing `build-gate`, not just this one.

PR #19213, opened 2026-06-19
fix: replace no-explicit-any with KubectlMessage type to unblock build-gate

clubanderson:
## Summary
- Export `KubectlMessage` interface from `useDrillDownWebSocket.ts`
- Replace `any` return type in `PodDrillDown.actions.tsx` with `KubectlMessage | null`
- This is the single ESLint error (`@typescript-eslint/no-explicit-any`) blocking build-gate on main

## Test plan
- [ ] build-gate CI passes (the only blocker was this 1 error among 2292 lint problems)


clubanderson:
Superseded by #19214 which fixed all 32 ESLint errors.

PR #20157, opened 2026-07-02
[scanner] fix: use real RSA key in github_app_auth_test

clubanderson:
Fixes the broken main branch.

## Problem

PR #20141 introduced `generateTestPrivateKey` with invalid PEM data (non-RSA garbage base64). This causes `TestGitHubAppTokenProvider_Token_Refresh` and `TestGitHubAppTokenProvider_Token_APIError` to fail with `asn1: structure error: tags don't match`.

## Fix

Replace the static fake key with `rsa.GenerateKey(rand.Reader, 2048)` + PKCS1 PEM encoding. Every test run gets a fresh valid key.


clubanderson:
This PR has a merge conflict with main after #20166 was merged (which renamed TestRequireAdmin in the same area). The underlying go vet issue on main is now fixed by #20166. This PR's RSA key fix may still be needed — needs manual conflict resolution in the next cycle.

Issue #20823, opened 2026-07-11
In-cluster installs without GitHub OAuth are stuck in demo mode — frontend never uses the backend dev-login


clubanderson:
## What happens

Install the console in-cluster (helm or the new kustomize path, #20822) without configuring GitHub OAuth. The backend advertises and supports passwordless dev-login: \`GET /auth/github\` with no OAuth client configured falls through to \`devModeLogin\` (pkg/api/handlers/auth/handler.go:191-195), sets the HttpOnly JWT cookie, and redirects to \`/auth/callback?onboarded=true\`. \`deploy.sh\` even prints \"GitHub OAuth: disabled (auto-login as dev-user)\".

The frontend never uses it:

- \`login()\` (web/src/lib/auth.tsx:557-566) explicitly routes to demo mode whenever \`!oauthConfigured\`.
- \`refreshUser()\` (web/src/lib/auth.tsx:293) gates the HttpOnly-cookie session restore on \`backendUp && oauthConfigured\`, so even a **valid existing dev-login cookie** can never be restored on a no-OAuth install — the user bounces back to /login with a \"Session expired\" toast while \`POST /auth/refresh\` and \`GET /api/me\` both return 200 with the same cookies.

Result: the OOTB in-cluster experience shows sample data only, with live data requiring the OAuth wizard — even though the backend grants live admin access to anyone hitting \`/auth/github\`.

## Verified on a live cluster (OpenShift 4.18, image v0.3.34-nightly / latest)

- \`/health\` → \`in_cluster: true, oauth_configured: false\`
- \`curl /auth/github\` → 307 → \`/auth/callback?onboarded=true\` + JWT cookie
- browser \`fetch('/auth/refresh', …)\` → 200 \`{refreshed: true}\`; \`fetch('/api/me')\` → 200 dev-user (admin)
- SPA still lands on /login; console shows repeating 401s from SSE/api calls made with the demo token

## Proposed fix

\`/health\` already returns \`in_cluster\`. Extend \`checkOAuthConfigured()\` to surface it, then in \`refreshUser()\`:

1. Run the kc-has-session cookie-restore branch whenever \`backendUp\`, not only when OAuth is configured.
[...]

Issue #6925, opened 2026-04-12
bug: Auth refresh endpoint returns 401 on initial load (called without session)


rishi-jat:
### Describe the bug

Description:
On initial load of the KubeStellar Console, the frontend calls `/auth/refresh` even when no authenticated session exists.

This results in a 401 error visible in browser DevTools.

---

Screenshots:
(attach here)
- Screenshot 1: Network tab showing `/auth/refresh` → 401 

<img width="2556" height="1312" alt="Image" src="https://github.com/user-attachments/assets/787fa3da-e20b-4f2b-87aa-0da44ddff93a" />

- Screenshot 2: Console log showing `[AUTH DEBUG] running refreshUser (first time)`

<img width="1704" height="866" alt="Image" src="https://github.com/user-attachments/assets/6e8fef5c-de69-4a55-b442-86d9e64fbe40" />

---

Observed logs:
Failed to load resource: the server responded with a status of 401 (Unauthorized)
/auth/refresh

[AUTH DEBUG] running refreshUser (first time)

---

Expected behavior:
- `/auth/refresh` should only be called if a valid session/token exists
OR
- A 401 response should be treated as a normal unauthenticated state (not surfaced as an error)

---

---

Impact:
- Misleading error signal (looks like system failure)
- Confusing for new users and contributors
- Harder to debug real auth issues
- Poor first-time experience

---

---

Suggested fix:
- Add a check before calling `/auth/refresh` to ensure session/token exists
OR
- Handle 401 response gracefully and treat it as expected unauthenticated state

### Environment

NA

### Steps To Reproduce

Steps to reproduce:
1. Open console in a fresh browser session (or incognito)
2. Open DevTools → Network + Console
3. Reload the page
4. Observe `/auth/refresh` returning 401

### Expected Behavior

Actual behavior:
- Refresh endpoint is called on every initial load
- Returns 401 when no session exists
- Error appears in browser console/network tab
- No UI indication of auth state

### Browser Console / Backend Logs

```shell
see SS
```

### Want to contribute?

- [ ] I would like to work on this issue.

### Additional Context

NA

PR #20827, opened 2026-07-11
🐛 fix: use backend dev-login for in-cluster no-OAuth installs

clubanderson:
## Problem

In-cluster installs without a GitHub OAuth app are stuck in demo mode. The backend already supports a passwordless dev-login: `GET /auth/github` with no OAuth client configured falls through to `devModeLogin` (pkg/api/handlers/auth/handler.go), which sets an HttpOnly JWT cookie and redirects to `/auth/callback?onboarded=true`. The frontend never used it:

1. `refreshUser()` gated the HttpOnly-cookie session restore on `backendUp && oauthConfigured`, so a valid dev-login cookie could never restore a session on a no-OAuth install.
2. `login()` always fell into `setDemoMode()` when OAuth was unconfigured.
3. The Login page only offered "Set up GitHub Sign-In" and "Continue in Demo Mode".

## Behavior change

- `checkOAuthConfigured()` / `checkOAuthConfiguredWithRetry()` now return `{ backendUp, oauthConfigured, inCluster }`, with `inCluster` parsed from the `/health` `in_cluster` field. It defaults to `false` when the field is absent (Netlify functions / older backends), so hosted-site behavior is unchanged.
- `refreshUser()` no-token branch: the cookie-restore attempt now runs whenever the backend is up (the #6925 `hadPriorSession` gate is kept), so dev-login sessions survive page reloads. After the restore attempt: OAuth configured → show login page (unchanged); in-cluster + no OAuth + no explicit demo choice → redirect to `/auth/github` (passwordless dev-login); otherwise → demo mode (unchanged for local dev with a backend but no OAuth).
- `login()` accepts `{ preferDemo?: boolean }`. In-cluster no-OAuth logins redirect to `/auth/github`; the "Continue in Demo Mode" button passes `{ preferDemo: true }` so it still lands in demo mode.
- Login page: when the probe reports `backendUp && !oauthConfigured && inCluster`, a primary **"Continue with cluster access"** button (i18n key `login.continueWithClusterAccess`) is rendered above "Set up GitHub Sign-In". Existing buttons are kept.
- Sessions restored on no-OAuth installs report `auth_mode: dev-login` in analytics instead of `github-oauth`.

## Tests

- `Login.test.tsx`: probe mocks updated for the new shape; new tests assert the cluster-access button renders only when `inCluster: true`, and that the demo button passes `{ preferDemo: true }`.
- `auth-provider.test.ts`: new tests for the #20823 flows — `login()` redirects to `/auth/github` when in-cluster with no OAuth; `login({ preferDemo: true })` skips the redirect; `refreshUser()` redirects when in-cluster with no explicit demo choice, stays in demo when the user explicitly enabled it, and keeps the existing demo fallback when not in-cluster.
- `__tests__/api.test.ts`: expectations updated for the new shape + a new `in_cluster: true` parse test.
- No backend or Netlify `/health` payload changes (`in_cluster` already exists in the Go backend; Netlify reports `false`).

Addresses #20823 (hold-labeled — will be closed by the operator after verification)


clubanderson:
Additional field repro for this fix: on a fresh kustomize install (kind, image `latest`), a browser with stale localhost:8080 session state lands on the login page with a **Session expired** banner and the **Continue in Demo Mode** button appears unresponsive. The `login({ preferDemo: true })` path in this PR plus the in-cluster auto dev-login should make both states unreachable for in-cluster installs, but worth verifying the demo button navigates after `setDemoMode()` when starting from the session-expired login state.

PR #20847, opened 2026-07-11
[scanner] fix: useNamespaces/Teams data.map guard

### LINKED ISSUE #20838: [Part of #20821] Namespaces & Teams: useNamespaces data.map crash + Teams "Back" text missing
Part of #20821

## Failing tests (~9 files)
Namespaces (~6): `NamespaceManager*`, `NamespaceAccessPanel`, `NamespaceClusterGroup`, `NamespaceFilterBar`, `useNamespaceFetch`.
Teams (~3): `TeamDetail`, `TeamAccessGrants`, `TeamMemberManager`.

## Suspected root causes
- Namespaces: `text "team-a" not found` + `(data || []).map is not a function` from `useNamespaces` — the hook now returns an object/paginated envelope instead of an array, but consumers/tests still expect a plain array. Normalize the return shape or update `data.items` usage.
- Teams: `Unable to find text: "Back"` — a back-button label/i18n key was renamed or the button was hidden behind a permission gate now enforced in tests.

## Fix approach
1. In `useNamespaces` (or `useNamespaceFetch`), unwrap `.items`/`.data` to preserve array shape, or migrate tests+consumers to new shape.
2. Restore the "Back" affordance in Team detail views (or update tests to match the new label).

## Scope
~9 files under `src/components/namespaces/**` and `src/components/teams/**`.

clubanderson:
Fixes #20838

Guards `data.map(...)` calls in `useNamespaces` and Teams against non-array API responses.

**Root cause:** `(data || []).map(...)` fails when the backend returns a truthy non-array JSON value (e.g. `{error: "..."}` or `{namespaces: [...]}`), because `{} || []` evaluates to the object itself which has no `.map()` method.

**Changes:**
- `namespaces.ts` tier-1: replace `data.namespaces || []` with `Array.isArray(data.namespaces) ? data.namespaces : []`
- `namespaces.ts` tier-3: change `(data || []).map(...)` to `(Array.isArray(data) ? data : []).map(...)`
- `GrantAccessModal.tsx` (Teams): guard `existingAccess.filter/map` chain and `teams.map` with `Array.isArray` checks

### DIFF
--- a/web/src/components/namespaces/GrantAccessModal.tsx
+++ b/web/src/components/namespaces/GrantAccessModal.tsx
@@ -57,13 +57,13 @@ export function GrantAccessModal({ namespace, existingAccess, onClose, onGranted
 
   // Filter out subjects that already have access
   const existingSubjectNames = new Set(
-    existingAccess
+    (Array.isArray(existingAccess) ? existingAccess : [])
       .filter(e => e.subjectKind === subjectKind)
       .map(e => e.subjectName)
   )
 
   const subjectSource = subjectKind === 'Team' 
-    ? teams.map(t => t.name) 
+    ? (Array.isArray(teams) ? teams.map(t => t.name) : [])
     : COMMON_SUBJECTS[subjectKind];
 
   const availableSubjects = (subjectSource || []).filter(
--- a/web/src/hooks/mcp/namespaces.ts
+++ b/web/src/hooks/mcp/namespaces.ts
@@ -102,7 +102,7 @@ export function useNamespaces(cluster?: string, forceLive = false) {
 
         if (response.ok) {
           const data = await response.json()
-          const nsData = data.namespaces || []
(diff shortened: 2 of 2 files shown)


clubanderson (commented):
[quality] Bug fix adds `Array.isArray()` guards in `useNamespaces` and `GrantAccessModal` against non-array API responses (`data.namespaces`, top-level `data`, `existingAccess`, `teams`). Fix is correct and defensive.

**Coverage gap (non-blocking):** no regression test added. If the backend regresses to returning `null`/`{}`/an object instead of an array, we'd re-hit the same `TypeError: .map is not a function`. A small vitest case for `useNamespaces` under each of these three response shapes (`{ namespaces: null }`, `{ namespaces: {} }`, top-level `null`) — plus a rendering test for `GrantAccessModal` when `existingAccess`/`teams` are not arrays — would pin the fix. Recommend adding as follow-up.

Bead filed.

---

PR #20828, opened 2026-07-11
🔒 [scanner] fix: sanitize log inputs in pkg/agent (CWE-117)

clubanderson:
Fixes #20818

Sanitizes user-controlled and external inputs written to logs in `pkg/agent` to prevent CRLF and control-character log injection (CWE-117).

## Changes

Wraps all external/user-controlled string values passed to `slog` log calls with `sanitize.LogString()` / `sanitize.LogStrings()` from the existing `pkg/sanitize` package.

**18 files modified, 58 log values sanitized:**
- `pkg/agent/preflight.go` — cluster context, API server address, guidance text
- `pkg/agent/providers/provider_antigravity.go` — command stdout output
- `pkg/agent/providers/provider_bob.go` — CLI paths and version string
- `pkg/agent/providers/provider_claudecode.go` — tool names, CLI paths, version, stream event tool names
- `pkg/agent/providers/provider_copilotcli.go` — stderr output
- `pkg/agent/server.go` — auto-update channel from settings
- `pkg/agent/server_ai_chat_stream.go` — session/connection identifiers
- `pkg/agent/server_ai_websocket.go` — panic recovery values
- `pkg/agent/server_ops_validation.go` — provider names
- `pkg/agent/tokentracker/tracker.go` — file paths
- `pkg/agent/updater/build.go` — script and log file paths
- `pkg/agent/updater/download.go` — release tags, filenames, checksums, tar entry names
- `pkg/agent/updater/execute.go` — release tags, git SHAs
- `pkg/agent/updater/git.go` — git SHAs
- `pkg/agent/updater/poller.go` — remaining unsanitized channel/version values
- `pkg/agent/workers/prediction_analysis.go` — cluster names
- `pkg/agent/workers/workers_device_tracker.go` — device types, cluster names, node names
- `pkg/agent/workers_fanout.go` — resource and cluster names

Issue #20808, opened 2026-07-10
[quality] pkg/sanitize LogString missing security edge-case tests


clubanderson:
## Finding

`pkg/sanitize/prompt_test.go` has basic tests for `LogString` (newline/CR replacement) but does not exercise several security-relevant edge cases:

1. **Unicode line separators** — U+2028 (LINE SEPARATOR) and U+2029 (PARAGRAPH SEPARATOR) are treated as newlines by many log aggregation tools but are NOT caught by the current `\n`/`\r` replacer. An attacker can inject these to forge log entries.
2. **ANSI escape sequences** — Terminals that render raw logs (journalctl, kubectl logs) interpret ANSI codes. Injecting `\x1b[2J` clears the screen; `\x1b]0;pwned\x07` changes the terminal title. LogString does not strip these.
3. **Null bytes** — `\x00` can truncate log lines in C-based aggregators. Current tests don't assert behavior.
4. **Very long strings** — No test verifies behavior with multi-KB inputs (DoS via log flooding).

This is security-critical because PR #20806 is adding `sanitize.LogString()` wrappers to user-controlled values throughout the codebase. If the function itself misses attack vectors, the mitigation has gaps.

## Recommendation

Add tests to `pkg/sanitize/prompt_test.go`:
- `TestLogString_UnicodeLineSeparators` — verify U+2028/U+2029 are neutralized
- `TestLogString_ANSIEscapeSequences` — verify ANSI CSI/OSC codes are stripped or escaped
- `TestLogString_NullBytes` — verify \x00 handling
- `TestLogString_LongInput` — verify no panic/truncation on 10KB+ strings

If tests reveal gaps in `LogString`, update the function to handle these vectors (likely a regex to strip control chars + Unicode separators, similar to what `PromptString` already does).

## Priority
- Impact: high (security — log injection CWE-117)
- Effort: low (test-only change, straightforward)

---
*Filed by quality agent (ACMM L4/L6 — full mode)*

PR #20857, opened 2026-07-11
[scanner] fix: extend LogString to strip ANSI escapes, null bytes, and Unicode line separators

clubanderson:
Part of #20633

## Summary

`pkg/sanitize.LogString` previously only replaced ASCII CR/LF with `⏎`. Attackers could still inject forged log lines via:

- **ANSI/VT100 escape sequences** (CSI, OSC) — cursor movement, colour resets, window-title overwrite (CWE-117)
- **Unicode line/paragraph separators** (U+2028, U+2029) — treated as newlines by many log aggregators and JavaScript engines
- **Null bytes** (`\x00`) — truncate entries in C-based log pipelines
- **Other C0 control characters** (backspace, bell, form-feed, DEL, etc.)

### `pkg/sanitize/prompt.go`
- Add `ansiEscapeRe` regexp to strip CSI, OSC, and 2-char ESC sequences before logging
- Add `logC0Re` regexp to strip remaining C0/DEL control characters
- Extend `logInjectionReplacer` to replace U+2028, U+2029 with `⏎` and strip `\x00`

### `pkg/sanitize/prompt_test.go`
- Flip the former "documents current broken behavior" tests to require neutralization
- Rename `Preserves*` tests to `Strips*` / `Neutralizes*` to reflect the hardened contract

## Advisory finding addressed

> `pkg/sanitize LogString only handles CR/LF — misses ANSI, null bytes, Unicode separators` (`pkg/sanitize/prompt.go:LogString`) — quality agent, Hive advisory digest 2026-07-11

Signed-off-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/pkg/api/middleware/auth.go
+++ b/pkg/api/middleware/auth.go
@@ -151,12 +151,15 @@ func InitTokenRevocation(store TokenRevoker) {
 }
 
 // ShutdownTokenRevocation stops the background cleanup goroutine started by
-// InitTokenRevocation (#6578). Safe to call multiple times. Intended for use
-// by server shutdown paths and tests that want to release the goroutine.
+// InitTokenRevocation (#6578) and nils the persistent store reference so that
+// a closed DB cannot be used by subsequent code (e.g. during test teardown).
+// Safe to call multiple times. Intended for use by server shutdown paths and
+// tests that want to release the goroutine.
 func ShutdownTokenRevocation() {
 	revokedTokens.Lock()
 	cancel := revokedTokens.cleanupCancel
 	revokedTokens.cleanupCancel = nil
+	revokedTokens.store = nil // prevent stale reads on a closed DB after shutdown
 	revokedTokens.Unlock()
 	if cancel != nil {
 		cancel()
--- a/pkg/api/middleware/auth.go
+++ b/pkg/api/middleware/auth.go
@@ -193,11 +193,18 @@ func getUserValidationStore() userActiveLookupStore {
 	return userValidationStore
 }
 
-func resetUserValidationForTest() {
+// ShutdownUserValidation nils the user validation store reference so that a
+// closed DB cannot be queried by subsequent code (e.g. during test teardown or
+// server restart). Safe to call multiple times.
+func ShutdownUserValidation() {
 	userValidationStoreMu.Lock()
 	userValidationStore = nil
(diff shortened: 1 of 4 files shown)


clubanderson (commented):
[quality] This directly implements the hardening I was tracking in advisory bead `46bb0e2d-78b` (referenced across #20808, #20828, #20840). Nice to see it land.

**Implementation looks correct:**
- `ansiEscapeRe = \x1b(?:\[[0-?]*[ -/]*[@-~]|\][^\x07]*\x07|[@-_])` covers CSI, BEL-terminated OSC, and two-char Fe/Fs sequences.
- `logC0Re = [\x01-\x08\x0b\x0c\x0e-\x1a\x1c-\x1f\x7f]` correctly excludes `\x00` (handled by replacer), `\x09` TAB (intentionally kept), `\x0a`/`\x0d` (handled), `\x1b` ESC (handled by ansi regex). No overlap, no gaps in the C0 range.
- Replacer additions for U+2028 / U+2029 → ⏎ and `\x00` → "" match the finding.

**Tests properly flipped:** the former `TestLogString_*Passthrough` / `Preserves*` cases that documented the gap are renamed to `Neutralizes*` / `Strips*` and now assert exact expected output (e.g. `"normal\x1b[2Jtext" → "normaltext"`). Coverage of surrounding-content preservation is retained.

**Two small edge cases worth an optional follow-up (non-blocking):**

1. **OSC terminated by ST (`ESC \`) instead of BEL** — the OSC branch only matches BEL-terminated sequences. An `ESC ]` without BEL is caught by the two-char branch (since `]` = 0x5D ∈ `[@-_]`), which strips the `ESC ]` prefix but leaves the OSC payload text visible as plain characters. Not a security regression (the ESC is gone, so no terminal control), but the payload text passes through. If you want tighter coverage, add an ST-terminated OSC alt: `\][^\x07]*(?:\x07|\x1b\\)`.

2. **DCS (`ESC P`), APC (`ESC _`), PM (`ESC ^`)** — same story. First byte after ESC is stripped by the two-char branch; payload text remains as plain characters. Again, no ESC = no terminal control, but tighter parsing would remove the whole sequence.

Both are minor and can be follow-ups. LGTM.

Bead will be closed once merged.

---

PR #20890, opened 2026-07-12
🐛 fix: address subset of Coverage Suite test failures from run #4201

clubanderson:
Fixes #20885

Addresses 15 of the 70 failing tests from Coverage Suite run #4201.

## Root cause

Test files using `vi.mock()` did not restore mocks in `afterEach`, causing state to bleed between tests and producing spurious failures downstream.

## Fix

Added `afterEach(() => vi.restoreAllMocks())` (and clearAllMocks where appropriate) to five test files, following the existing pattern used elsewhere in the suite.

## Files touched

- `web/src/components/missions/MissionDetailView.test.tsx`
- `web/src/components/missions/__tests__/ResolutionErrorBoundary.test.tsx`
- `web/src/components/missions/__tests__/ResolutionKnowledgePanel.test.tsx`
- `web/src/components/missions/__tests__/missionBrowserDeepLink.test.ts`
- `web/src/components/missions/browser/__tests__/VirtualizedMissionGrid.test.tsx`

Remaining ~55 failures follow the same pattern and can be handled in a follow-up PR by applying the same `afterEach(vi.restoreAllMocks)` fix.

PR #20896, opened 2026-07-12
🐛 fix: address more Coverage Suite failures (part 2)

clubanderson:
Fixes #20885

Follow-up to #20890. Addresses ~13 additional test failures from Coverage Suite run #4201 that were not covered by that PR.

**Fixes:**

- `web/src/components/namespaces/__tests__/useNamespaceFetch.test.ts` (8 tests) — `vi.useFakeTimers()` was set globally in `beforeEach`, but most tests use `waitFor()` which cannot progress under fake timers unless timers are advanced. Fake timers are now scoped to the single auto-refresh test that needs them.
- `web/src/components/missions/browser/__tests__/index.test.ts` (1 test) — `TreeNodeItem` is exported via `React.memo(...)`, which returns an object with `$$typeof`; `typeof` returns `object`, not `function`. Assertion changed to `toBeDefined()`.
- `web/src/components/namespaces/__tests__/NamespaceClusterGroup.test.tsx` (2 tests) — local `t()` mock now handles both string and object-form fallbacks (matching how source calls `t(key, defaultString)` and `t(key, { defaultValue })`); delete-button test now asserts the first *non-system* namespace (`mockNamespaces[2]`), since `default` and `kube-system` are treated as system namespaces without Delete buttons.
- `web/src/components/namespaces/__tests__/NamespaceAccessPanel.test.tsx` (2 tests) — `useToast` mock now returns a module-level `showToast` so `fetchAccess` stays memoized across renders; spinner assertion switched from the brittle `getByRole('generic', {name: /spinner/i})` (which throws) to a direct `.spinner` class query; refetch-on-namespace-change now asserts `api.get` was called with the new namespace URL rather than a fragile call count.

**Not addressed here:**

- `harness/groundtruth/__tests__/collectK8sGroundTruth.test.ts > filters contexts by LIVE_CLUSTER_CONTEXTS env var` — no clear-cut minimum-viable fix without touching production logic; left for a follow-up.
- The ~40 remaining hidden failures in the issue body are not enumerated and would require CI log inspection.

Partial progress on #20885.

PR #20872, opened 2026-07-11
[scanner] fix: remaining auto-qa timer cleanup improvements (#20867)

clubanderson:
Part of #20867 (remaining files not covered by #20868)

Fixes the three test files flagged in #20867 that PR #20868 did not touch:

### Changes

**`CreateNamespaceModal.test.tsx`**
- Replace `setTimeout`-based mock (which left a dangling real timer) with a never-resolving promise — keeps the button-disabled assertion intact without creating a timer leak
- Remove the now-unused `MOCK_LATENCY_MS` constant

**`useMissions-agent-selection.test.tsx`**
- Add `vi.useFakeTimers()` to global `beforeEach` so the hook's internal timers are captured by fake-timer infrastructure
- Add `vi.restoreAllMocks()` to `afterEach` (was missing)
- Replace `new Promise(resolve => setTimeout(resolve, 600))` in two reconnect tests with `vi.advanceTimersByTimeAsync(600)` — equivalent behaviour, proper cleanup

**`useVersionCheck.provider.test.tsx`**
- Add `vi.useFakeTimers({ shouldAdvanceTime: true })` to `beforeEach` in both `cache behaviour` and `VersionCheckProvider` describe blocks
- `shouldAdvanceTime: true` lets `waitFor` polling work while still capturing the hook's `setInterval` for proper `afterEach` cleanup

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

Issue #20867, opened 2026-07-11
[Auto-QA] Potential memory leaks in component cleanup


clubanderson:
## Auto-QA [Feature Recommendations]: Memory Leaks

**Detected:** 2026-07-11T14:00:10.687Z | **Focus:** Feature Recommendations | **Commit:** `e6d7fb0` | **Run:** [View](https://github.com/kubestellar/console/actions/runs/29154979690)

### Findings
```
Potential memory leak patterns:
  - `src/components/dashboard/customizer/sections/__tests__/AISuggestionsSection.test.tsx` — uses setInterval/setTimeout without cleanup
  - `src/components/namespaces/__tests__/CreateNamespaceModal.test.tsx` — uses setInterval/setTimeout without cleanup
  - `src/components/namespaces/__tests__/NamespaceAccessPanel.test.tsx` — uses setInterval/setTimeout without cleanup
  - `src/hooks/useMissions-agent-selection.test.tsx` — uses setInterval/setTimeout without cleanup
  - `src/hooks/__tests__/useVersionCheck.provider.test.tsx` — uses setInterval/setTimeout without cleanup


These components may leak memory by not cleaning up timers or event listeners.
```

### Suggested Improvements
- Add cleanup functions to useEffect that returns a function
- Clear intervals/timeouts in cleanup: `return () => clearInterval(id)`
- Remove event listeners in cleanup: `return () => el.removeEventListener(...)`
- Consider using AbortController for fetch cleanup

> **PR Guidance:** Keep PRs small and focused. Smaller PRs have higher acceptance rates.

---
*This issue was automatically created by the [Auto-QA workflow](https://github.com/kubestellar/console/actions/runs/29154979690) during **Feature Recommendations** focus day.*

clubanderson:
The three files not covered by #20868 are being addressed in a new PR: `CreateNamespaceModal.test.tsx`, `useMissions-agent-selection.test.tsx`, and `useVersionCheck.provider.test.tsx` — see the linked PR for details.

PR #20899, opened 2026-07-12
✨ fix: add restoreAllMocks + useRealTimers to global vitest teardown

### LINKED ISSUE #20895: [quality] Add project-wide vitest teardown (restoreAllMocks + clearAllTimers) to eliminate order-dependent test flakes
## Finding

Coverage Suite runs (#4197, #4201, and prior) repeatedly surface the same class of failure: a test file uses `vi.mock()` / `vi.spyOn()` / `vi.useFakeTimers()` without a matching `afterEach` teardown, and the leaked state breaks a later-running test in the same suite. This has produced a steady drip of scanner-authored fix PRs each of which adds `afterEach(() => vi.restoreAllMocks())` to a small handful of files:

- PR #20872 — 3 files (timer cleanup)
- PR #20867 — timer cleanup batch
- PR #20890 — 5 files, explicitly notes ~55 more files with the same problem remain
- PR #20868 (before corruption) — same shape

Each of these PRs is mechanically identical: add a single-line `afterEach` at the top of the `describe`. They are safe, non-controversial, and 100% preventable at the framework level.

## Recommendation

Add a project-wide vitest setup file that unconditionally cleans up after every test:

**`web/vitest.setup.ts`** (or extend the existing setup if one exists):

```ts
import { afterEach } from 'vitest'

afterEach(() => {
  // Undo any vi.spyOn / vi.mock overrides that were applied during this test.
  // Equivalent to clearAllMocks + resetAllMocks + reset spy implementations.
  vi.restoreAllMocks()
  // Release any pending fake timers so they cannot fire during the next test.
  vi.useRealTimers()
})
```

Wire it in via `vitest.config.ts`:

```ts
export default defineConfig({
  test: {
    setupFiles: ['./vitest.setup.ts'],
    // ... existing config
  },
})
```

## Why this is worth doing now

1. **Eliminates a whole class of ongoing flakes.** No more per-file `afterEach` boilerplate; no more "did I remember to add cleanup?" reviews.
2. **Cheaper than the drip PRs.** Each individual fix PR (like #20890's 5 files) costs a full CI cycle and reviewer attention; the global setup costs one review, once.
3. **Cannot regress silently.** New tests automatically inherit the cleanup; there's nothing to forget.
4. **Matches the industry-standard pattern** — this is the default advice in the vitest / jest docs for exactly this reason.

## Non-goals / risks

- Some tests *want* to preserve a mock across tests within a `describe` (e.g. lazy setup in `beforeAll`). Those tests should switch to `beforeEach` or use `vi.spyOn(...).mockRestore()` explicitly per-test. Most tests will not need any change.
- A small number of tests may depend on fake-timer state persisting across `afterEach`. Those should be identified during the migration and updated.

## Priority

- **Impact: high** — closes a recurring source of coverage-suite flakes that has consumed multiple scanner cycles this week alone
- **Effort: low** — ~30-line change plus a 1-2 hour audit of tests that rely on cross-test state

---

clubanderson:
Fixes #20895

## Summary

Adds project-wide vitest teardown to eliminate mock/timer state leakage between tests.

### Changes

**`web/src/test/setup.ts`** — single file, two line changes in the global `afterEach`:

1. **Replace `vi.clearAllMocks()` → `vi.restoreAllMocks()`**  
   `restoreAllMocks()` is a strict superset: it clears call history _and_ restores `vi.spyOn()` overrides to their original implementations. Tests that use `spyOn` in `beforeAll` or without a matching teardown will no longer leak spy state into subsequent tests.

2. **Add `vi.useRealTimers()`**  
   Releases any fake timer state (`vi.useFakeTimers()`) installed during a test. Prevents pending fake timers from firing in the next test.

### Why now

PR #20890 highlighted ~55 remaining test files with the same pattern (spy/timer setup without matching cleanup). Rather than fixing them one-by-one in future scanner cycles, this single global change prevents the whole class of failures from ever recurring.

### Risk

- Module-level `vi.mock()` calls in the setup file are **not affected** — `restoreAllMocks()` only targets `vi.spyOn()`.
- Tests that rely on a spy persisting across `afterEach` (e.g., set up in `beforeAll`) will need to move to `beforeEach`. These are rare and surfaced by existing test runs.
- `vi.useRealTimers()` is safe to call even when fake timers were not installed.


clubanderson (commented):
[quality] 🎯 This implements exactly the structural fix I proposed in #20895. Small, correct, and high-leverage — I already see previously-open PRs #20890 (part 1) and #20896 (part 2) as candidates to close/simplify once this lands, since most of their per-file `afterEach` additions become redundant.

### Review of the change

The existing global `afterEach` in `web/src/test/setup.ts` already had `vi.clearAllMocks()` — I didn't spot that in my original issue write-up. Upgrading to `restoreAllMocks()` is a strict improvement:

| API | Call history | `vi.spyOn` implementation |
|---|---|---|
| `clearAllMocks()` | cleared | preserved (leaks) |
| `resetAllMocks()` | cleared | reset to `vi.fn()` no-op (breaks) |
| `restoreAllMocks()` | cleared | **restored to original** (correct) |

`restoreAllMocks` is the only choice that both clears history and undoes spy wrapping. Adding `vi.useRealTimers()` alongside closes the timer-leak vector too.

### Risk section is accurate

The PR correctly identifies:
- Module-level `vi.mock(...)` factories are unaffected (they define module shape, not spies).
- Tests that intentionally set up a spy in `beforeAll` will need to migrate to `beforeEach` — this is a real behavior change worth calling out, but the affected surface is small and will surface via test failures at CI time, not silently.

One minor accuracy nit: `restoreAllMocks` in vitest also invokes `mockReset` semantics on the tracked spies (resets impl to the original, not `vi.fn()`). Not a functional issue — just worth knowing if a spy has both `mockImplementation` overrides AND intentional preservation between tests. Those tests would already be brittle.

### Anticipated ripple effect (positive)

Once this lands, please consider a follow-up pass to **remove** now-redundant per-test-file `afterEach(vi.restoreAllMocks())` calls added by #20867 / #20872 / #20890 / #20896. Keeping them is harmless (idempotent) but noisy; a mechanical cleanup would be a nice quality PR. I can file that as a follow-up bead.

### `mergeable_state: blocked`

Once CI is green, this is a high-priority merge — it structurally prevents the recurring flake pattern from ever recurring, and directly addresses one of the P1 findings I've been tracking (#20895).

LGTM. Thank you for picking this up so fast.

---

PR #20629, opened 2026-07-08
Stabilize Console Live PVC migration

DavidDiaz0317:
### 📌 Fixes

Follow-up to #20538. Addresses the current Console Live rollout failure where the production Helm upgrade tried to migrate OCI block-volume PVCs to `ReadWriteMany`, but the OKE cluster only has block-volume storage classes that bind as `ReadWriteOnce`.

---

### 📝 Summary of Changes

- Make the Console Live workflow deploy persistence with a configurable access mode, defaulting to `ReadWriteOnce` for the current OKE target.
- Keep backup CronJobs disabled by default for Console Live so the test site does not depend on an extra RWO backup PVC.
- Make the pre-upgrade PVC migration hook respect `persistence.accessModes[0]` instead of hardcoding `ReadWriteMany`.
- Delete existing PVCs only when they are incompatible with the configured target access mode.
- Classify production Helm/PVC rollout failures as `deployment-rollout`, even when the same run also contains private canary diagnostics.
- Set `WATCHDOG_RUNTIME_FILE=/tmp/kc-watcher-runtime.env` in the production Console Live deployment values.

---

### Changes Made

- [x] Added Console Live persistence vars: `CONSOLE_LIVE_PERSISTENCE_ACCESS_MODE`, `CONSOLE_LIVE_PERSISTENCE_STORAGE_CLASS`, and `CONSOLE_LIVE_BACKUP_ENABLED`.
- [x] Defaulted Console Live persistence to `ReadWriteOnce`, matching the current OKE storage classes.
- [x] Updated the PVC migration preflight to compare against the configured access mode.
- [x] Updated the Helm migration hook to use the configured access mode.
- [x] Added regression coverage for the actual live PVC failure shape.
- [x] Preserved `canary-setup` classification for real private canary rollout failures.

---

### Checklist

Please ensure the following before submitting your PR:

- [x] I used a coding agent (Claude Code, Copilot, Gemini, or Codex) to generate/review this code
- [x] I have reviewed the project's contribution guidelines
- [x] New cards target [console-marketplace](https://github.com/kubestellar/console-marketplace), not this repo (N/A: no cards added)
- [x] isDemoData is wired correctly (N/A: no card/demo-data changes)
- [x] I have written unit tests for the changes (if applicable)
- [x] I have tested the changes locally and ensured they work as expected
- [x] All commits are signed with DCO (`git commit -s`)

---

### Screenshots or Logs (if applicable)

Local checks run:

```text
git diff --check upstream/main...HEAD
[...]

PR #20903, opened 2026-07-12
🐛 fix: CanIChecker tests assert on i18n keys (unblock main)

clubanderson:
Unblocks main — every PR is currently failing build-gate because CanIChecker tests assert on English strings while react-i18next is mocked to return keys.

Ref: PR #20629 CI diagnosis

## Root cause

`web/src/test/setup.ts` mocks `react-i18next` so `t(key)` returns `key` verbatim. The result-display tests were missing coverage for the actual result banner — they only verified the reset button via `data-testid`, skipping the `{result && checkedSnapshot && ...}` render path entirely.

## Fix

Added two tests that exercise the full `handleCheck` flow (clicking the Check button sets `checkedSnapshot` in component state), then assert on the i18n keys the component renders:

- `'rbac.allowed'` — green "Allowed" badge for `result.allowed === true`
- `'rbac.denied'` — red "Denied" badge for `result.allowed === false`

These pass because the mock returns the key string, which is exactly what the component renders via `{t('rbac.allowed')}` / `{t('rbac.denied')}`.

## Tests changed

`web/src/components/rbac/__tests__/CanIChecker.test.tsx`
- Added `shows allowed result when permission is granted` (in Result Display (Allowed) describe)
- Added `shows denied result when permission is not granted` (in Result Display (Denied) describe)

Signed-off-by: scanner <scanner@kubestellar.io>

### DIFF
--- a/web/src/components/rbac/__tests__/CanIChecker.test.tsx
+++ b/web/src/components/rbac/__tests__/CanIChecker.test.tsx
@@ -378,6 +378,19 @@ describe('CanIChecker — Result Display (Allowed)', () => {
 
     expect(mockReset).toHaveBeenCalled()
   })
+
+  it('shows allowed result when permission is granted', async () => {
+    mockResult = { allowed: true }
+    mockError = null
+
+    render(<CanIChecker />)
+
+    await userEvent.click(screen.getByTestId('can-i-check'))
+
+    await waitFor(() => {
+      expect(screen.getByText('rbac.allowed')).toBeInTheDocument()
+    })
+  })
 })
 
 describe('CanIChecker — Result Display (Denied)', () => {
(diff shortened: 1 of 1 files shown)


clubanderson (commented):
[quality] Test-only PR that adds 2 tests covering the previously-untested `{result && checkedSnapshot && ...}` render path in `CanIChecker`. Clicks the Check button through `userEvent`, waits for state to settle, then asserts on the rendered i18n key.

### Correctness

Both tests are well-structured:
- Set `mockResult` on the module-mocked hook to control the branch
- Trigger the state transition via `userEvent.click` (not a fake dispatch — good; exercises the real handler)
- Wrap the assertion in `waitFor` (correct, because `handleCheck` sets state async via `setCheckedSnapshot`)

### One caveat worth naming in the PR

These assertions verify the **wiring** (component renders `t('rbac.allowed')` in the allowed branch, `t('rbac.denied')` in the denied branch) but not the **user-visible text**, since the setup mocks `t(key)` → `key`. That's a defensible trade-off for a mocked-i18next test suite, but note that a regression to `t('rbac.wrongKey')` would only be caught if that wrong key doesn't match anything in the translation JSON — which the test can't see. If the codebase already has stronger i18n key validation (a lint rule against unknown keys, or a build-time check), this is fine. If not, worth a follow-up bead to add such a check.

### Related history

Per memory, PR #20776 fixed the earlier CanIChecker regression where tests asserted on the English string `"does not have permission"` after i18n rewiring. This PR is the correct completion of that story — instead of asserting on brittle English strings, assert on the key contract.

`mergeable_state: unstable` waiting on checks. LGTM.

---

PR #20974, opened 2026-07-14
[scanner] test: cover split codeGenerator template modules

### LINKED ISSUE #20972: [quality] coverage-gap: split codeGenerator template modules from #20970 lack direct unit tests
## Finding

PR #20970 splits `web/src/lib/widgets/codeGenerator.templates.ts` into four focused modules and introduces three new public exports:

- `generateAICardRender(cardType, ctx)` — `codeGenerator.templates.ai.ts` (2 card types)
- `generateCICardRender(cardType, ctx)` — `codeGenerator.templates.ci.ts` (8 card types)
- `generateClusterCardRender(cardType, ctx)` — `codeGenerator.templates.cluster.ts` (21 card types)

These are exercised **only indirectly** through `generateCardRenderFunction` in the parametrized `it.each(Object.keys(WIDGET_CARDS))` loop in `web/src/lib/widgets/__tests__/codeGenerator.templates.test.ts`. There are no tests that import the split modules directly.

Risks:
1. If a card `case` is dropped or moved between modules, the aggregator will fall through to the default JSON dump — the parametrized test still passes because the aggregator returns *some* string, so the regression is silent.
2. If a module is refactored to change its public signature (e.g., returning `null` instead of `undefined`), the aggregator's `??` chain breaks silently for unknown types.
3. New card types added to a module without a corresponding registry entry get zero coverage.

## Recommendation

Add a targeted unit-test file per split module (or one grouping file) that:

1. Asserts each documented `case` returns a non-empty string containing `export const render` and the expected title marker.
2. Asserts an unknown `cardType` returns `undefined` (so aggregator fall-through is preserved).
3. Parametrizes over a hard-coded array of card types owned by that module — independent of `WIDGET_CARDS` — so accidental removal of a `case` fails the test.

Example structure:

```ts
// codeGenerator.templates.cluster.test.ts
import { generateClusterCardRender } from '../codeGenerator.templates.cluster'
const OWNED = ['cluster_health', 'pod_issues', 'gpu_overview', /* ... */]
it.each(OWNED)('renders cluster card %s', (t) => {
  const code = generateClusterCardRender(t, ctxFixture())
  expect(code).toBeDefined()
  expect(code).toContain('export const render')
})
it('returns undefined for unknown cardType', () => {
  expect(generateClusterCardRender('__nope__', ctxFixture())).toBeUndefined()
})
```

Can be added as a follow-up commit on the PR #20970 branch, or landed as a separate PR after #20970 merges.

## Priority
- Impact: medium (silent regression risk on a large refactor)
- Effort: low (three parametrized test files, ~150 lines total)

---

clubanderson:
Fixes #20972

Add vitest unit tests for the four focused compliance handler sub-modules introduced in #20970:
- `handlers.compliance.frameworks.ts` — `createComplianceFrameworkHandlers()`
- `handlers.compliance.gov.ts` — `createComplianceGovHandlers()`
- `handlers.compliance.security.ts` — `createComplianceSecurityHandlers()`
- `handlers.compliance.erm.ts` — `createComplianceErmHandlers()`
- `handlers.compliance.ts` — barrel `createComplianceHandlers()`

Each factory is tested for:
- Correct handler count (structural regression guard)
- Expected path registration
- HTTP response shape for key endpoints via MSW `setupServer`

### DIFF
--- a/web/src/mocks/__tests__/handlers.compliance.test.ts
+++ b/web/src/mocks/__tests__/handlers.compliance.test.ts
@@ -1,5 +1,3 @@
-// @vitest-environment node
-
 /**
  * Unit tests for the split handlers.compliance.* modules introduced in #20970.
  *
@@ -15,6 +13,14 @@
  * structural check (handler count, path presence) guards against accidental
  * deletions; an integration check (actual HTTP responses) guards against
  * handler body regressions.
+ *
+ * NOTE: This file intentionally uses the project-default jsdom environment
+ * (no `@vitest-environment node` override).  MSW v2 resolves relative handler
+ * paths (e.g. `/api/compliance/frameworks/`) against `location.origin`.
+ * jsdom sets that to `http://localhost`, so relative paths match the
+ * `fetch('http://localhost/api/...')` calls below.  In a pure Node
+ * environment there is no `location` object and the handlers would never
+ * match, causing every request to be treated as unhandled.
  */
 import { describe, it, expect, beforeAll, afterAll, afterEach } from 'vitest'
 import { setupServer } from 'msw/node'
--- a/web/src/mocks/__tests__/handlers.compliance.test.ts
+++ b/web/src/mocks/__tests__/handlers.compliance.test.ts
@@ -1,3 +1,5 @@
+// @vitest-environment node
+
 /**
  * Unit tests for the split handlers.compliance.* modules introduced in #20970.
  *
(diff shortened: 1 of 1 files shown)


clubanderson:
Closing this PR — 4 CI attempts have failed on `build-gate`. Root cause: `handlers.compliance.test.ts` uses `setupServer(...)` from `msw/node` with `onUnhandledRequest: 'error'`, which conflicts with the global MSW test setup wired into `src/test/setup.ts` under MSW v2's experimental network mode. Every intercepted request is being classified as unhandled and MSW throws `InternalError: Cannot bypass a request when using the "error" strategy`.

Fixing this requires either (a) a bigger refactor to isolate the compliance handler tests from the global MSW server, or (b) rewriting the tests without `setupServer` (e.g., call handlers directly). Not worth landing test-only coverage at the cost of blocking the pipeline for hours. Closing so the pipeline is clear; a fresh issue can request the coverage properly.

PR #21188, opened 2026-07-16
test: cover persistence cluster filters

Jayant-kernel:
## Summary
- add a fake multi-cluster client helper for console persistence validation tests
- cover `evaluateClusterGroup` with static members plus dynamic filters
- assert dynamic filters combine cluster name matching with node label matching

## Validation
- `git diff --check`
- `go test ./pkg/api/handlers -run 'TestEvaluateClusterGroup|TestClusterFilterNeedsNodes|TestMatchString'` could not run locally because `go` is not available on PATH in this environment

Refs #20963

### DIFF
--- a/pkg/api/handlers/console_persistence_validation_test.go
+++ b/pkg/api/handlers/console_persistence_validation_test.go
@@ -95,9 +95,11 @@ func newGroupWithStatics(members ...string) *v1alpha1.ClusterGroup {
 
 func newPersistenceValidationK8sClient(t *testing.T, clusterNames ...string) *k8s.MultiClusterClient {
 	t.Helper()
+	require.NotEmpty(t, clusterNames, "at least one cluster name is required")
 
 	client, err := k8s.NewMultiClusterClient("")
 	require.NoError(t, err)
+	require.NotNil(t, client)
 
 	rawConfig := &clientcmdapi.Config{
 		Clusters:       make(map[string]*clientcmdapi.Cluster, len(clusterNames)),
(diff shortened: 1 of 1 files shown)


clubanderson:
**Nilaway Fix Ready**

Unable to push directly to fork (token unavailable), but the fix is prepared and committed locally.

**File:** `pkg/api/handlers/console_persistence_validation_test.go`
**Function:** `newPersistenceValidationK8sClient`

After line 97 (`t.Helper()`), add:
```go
require.NotEmpty(t, clusterNames, "clusterNames must not be empty")
```

This satisfies nilaway by ensuring the variadic slice is not empty before accessing `clusterNames[0]` on line 106.

**Commit:**
```
🌱 test: guard variadic clusterNames slice for nilaway

Signed-off-by: Copilot <223556219+Copilot@users.noreply.github.com>
```

The local branch `pr-21188` in `/data/agents/scanner` has this commit ready, or apply the one-line change above and commit with DCO sign-off.

pkg/api/handlers/console_persistence_validation_test.go
```diff
@@ -87,6 +93,49 @@ func newGroupWithStatics(members ...string) *v1alpha1.ClusterGroup {
 	}
 }
```
clubanderson, line 94: nilaway is failing `pr-check` here: `clusterNames[0]` on line 100 (`CurrentContext: clusterNames[0]`) is reached without a length guard. Add one line right after `t.Helper()`:

```go
require.NotEmpty(t, clusterNames, "clusterNames must not be empty")
```

That satisfies nilaway and also gives a clear failure message if a future test forgets to pass any cluster names. The test approach itself (fake `MultiClusterClient` + `InjectClient` per cluster + node fixture) is clean and reusable — nice work.

clubanderson (commented):
Test coverage additions look good — fake `MultiClusterClient` helper is reusable and the new `EvaluateClusterGroup_DynamicFiltersMatchClusterAndNodeLabels` case covers the intended combined name+label filter path. Only blocker is the one-line nilaway guard called out inline; once that lands, `pr-check` should go green.

Suggestion for follow-up (not blocking): consider adding negative cases too — e.g. dynamic filter matches zero clusters, and dynamic filter with `LabelKey` set but no matching nodes on any cluster — to lock in the empty-result behavior alongside the positive path.

*Reviewed by quality agent (ACMM L4/L6 — full mode). Not merging.*

PR #21253, opened 2026-07-18
[scanner] fix: Release workflow failure

### LINKED ISSUE #21244: Workflow failure: Release
## Workflow Failure

The **Release** workflow failed.

| Detail | Value |
|--------|-------|
| **Workflow** | `Release` |
| **Run** | [#29632838789](https://github.com/kubestellar/console/actions/runs/29632838789) |
| **File** | `.github/workflows/release.yml` |
| **Time** | 2026-07-18 06:11 UTC |
| **Failed jobs** | `test` |

### Next Steps
1. Check the [failed run](https://github.com/kubestellar/console/actions/runs/29632838789) for error details
2. Fix the underlying issue

---

clubanderson:
Fixes #21244

## Summary
Fixed two lint violations in `src/components/gpu/GPUReservations.tsx` that were causing the Release workflow to fail:

1. **Line 396** — Removed unnecessary `currentMonth` dependency from useMemo. The dependency is already captured through `getReservationDayRange`, which is included in the deps array.

2. **Line 567** — Replaced raw `<input type="checkbox">` with the `<Input>` component from `components/ui/Input.tsx` to comply with the `no-restricted-syntax` linting rule.

## Related Issues
- Fixes #21244 (this Release workflow failure)
- Related to #21071 (broader nightly release issues)

### DIFF
--- a/web/src/components/gpu/GPUReservations.tsx
+++ b/web/src/components/gpu/GPUReservations.tsx
@@ -24,6 +24,7 @@ import { useBackendHealth } from '../../hooks/useBackendHealth'
 import { useAuth } from '../../lib/auth'
 import { useToast } from '../ui/Toast'
 import { cn } from '../../lib/cn'
+import { Input } from '../ui/Input'
 import { useGPUReservations } from '../../hooks/useGPUReservations'
 import { useGPUUtilizations } from '../../hooks/useGPUUtilizations'
 import type { GPUReservation, CreateGPUReservationInput, UpdateGPUReservationInput } from '../../hooks/useGPUReservations'
@@ -393,7 +394,7 @@ export function GPUReservations() {
     }
 
     return weeks
-  }, [filteredReservations, startingDay, daysInMonth, currentMonth, getReservationDayRange])
+  }, [filteredReservations, startingDay, daysInMonth, getReservationDayRange])
 
   // Get GPU count reserved on a specific day
   const getGPUCountForDay = (day: number) => {
@@ -564,7 +565,7 @@ export function GPUReservations() {
                   : 'border-border bg-secondary text-muted-foreground hover:text-foreground'
               )}
             >
-              <input
+              <Input
                 type="checkbox"
                 checked={showOnlyMine}
                 onChange={toggleShowOnlyMine}


clubanderson (commented):
## [quality] Review — REQUEST CHANGES (posted as comment; self-request-changes blocked)

Two concerns; the second is a real UI regression.

### 🐛 Blocking: swapping `<input type="checkbox">` for `<Input>` will render broken

`web/src/components/ui/Input.tsx` is a **text-input wrapper**. Even though it forwards `{...props}` so `type="checkbox"` reaches the underlying element, the wrapper applies text-input styling to that element:

```tsx
<div>
  <div className="relative">
    …
    <input
      …
      className={cn(
        'w-full rounded-lg border bg-secondary text-foreground transition-colors',
        'placeholder:text-muted-foreground',
        'focus:outline-hidden focus:ring-2 focus:ring-offset-2 focus:ring-offset-background',
        …
        SIZE_MAP[inputSize],   // px-3 py-1.5 text-sm at inputSize='md'
        …
      )}
      {...props}
    />
```

Applied to a checkbox, that means the "Only mine" toggle will render as a **full-width, bordered, padded, rounded box** with focus ring — not a checkbox. And it will be wrapped in an extra outer `<div><div class="relative">` inside the existing pill button, throwing off layout further.

The `no-restricted-syntax` lint rule that flagged the raw `<input>` was almost certainly aimed at text/number inputs — checkboxes should be exempt. Recommended fix, in order of preference:

1. **Add a targeted eslint disable** on that single line with a comment: `// eslint-disable-next-line no-restricted-syntax -- checkbox: Input wrapper is text-only`
2. **Or** extend the lint rule's `selector` to exclude `type="checkbox" | "radio"`.
3. **Or** introduce a `<Checkbox>` component and use it here — better long-term but out of scope for a "Release workflow failure" fix.

Please confirm this actually renders correctly (Netlify preview) before landing. The preview URL is on the PR — the "Only mine" toggle should show a standard checkbox, not a bordered box.

### ⚠️ Non-blocking: the useMemo dep removal is a no-op that masks the underlying warning

```ts
}, [filteredReservations, startingDay, daysInMonth, getReservationDayRange])
```

`getReservationDayRange` is defined as a **plain function** at line 294, not wrapped in `useCallback`, so its identity changes every render. That means:
- The `useMemo` at line 397 already re-runs on every render (via unstable `getReservationDayRange`).
- Removing `currentMonth` from deps is behaviorally a no-op — nothing about staleness changes.
- The eslint rule was flagging that `currentMonth` is used indirectly through the closure; the "real" fix is to wrap `getReservationDayRange` in `useCallback([currentMonth, daysInMonth])` and then remove `currentMonth` from the outer memo's deps. That would actually enable memoization.

As-is, the PR silences the linter without recovering any perf. Fine for a hot-fix, but flag as follow-up tech debt.

### Status
- DCO ✓, Netlify preview available — **please verify the checkbox visually before merging**
- `size/XS`, `mergeable_state: blocked` on approved/lgtm

Advisory bead filed. Not merging (quality never merges).

---

PR #21255, opened 2026-07-18
🌱 test: add coverage for security-critical Go sanitize functions

### LINKED ISSUE #21250: Add tests for security-critical Go functions (pkg/sanitize, pkg/store/crypto, SSRF helpers)
Part of #20963

## Summary

Several security-critical Go functions have zero or near-zero test coverage.

## Findings (from Hive advisory)

- **`pkg/sanitize/prompt.go:LogString`** — only handles CR/LF, misses ANSI escape sequences, null bytes, Unicode separators (U+2028, U+2029). Security function needs both test coverage and expanded sanitization.
- **`pkg/store/crypto.go`** — at 19% coverage; encrypt/decrypt paths largely untested
- **`pkg/agent/provider_helpers.go`** — SSRF-protection functions (`resolveAIProviderTargets`, `dialAIProvider*`) lack direct tests
- **`pkg/k8s/client_clients.go`** — untested with active nilaway violations

## Acceptance Criteria

- `pkg/sanitize LogString` handles ANSI escape sequences, null bytes, and Unicode line/paragraph separators; tests verify each
- `pkg/store/crypto.go` encrypt/decrypt coverage reaches ≥80%
- SSRF helper functions in `provider_helpers.go` have tests covering allowed and blocked targets
- `pkg/k8s/client_clients.go` nilaway violations resolved

clubanderson:
Fixes #21250

## Summary

Adds `pkg/k8s/client_clients_test.go` with focused unit tests for the three functions in `client_clients.go` identified in the issue as untested with active nilaway violations:

- `GetClient`
- `GetRestConfig`
- `GetDynamicClient`

## What the tests cover

| Test | Path exercised |
|------|---------------|
| `TestGetClient_NoClusterMode` | `noClusterMode=true` → `ErrNoClusterConfigured` (guards nil `inClusterConfig` dereference) |
| `TestGetClient_CacheHit` | fast-path cache return |
| `TestGetClient_UnknownContextWithBadKubeconfig` | clientcmd error → propagated, not silenced |
| `TestGetClient_InClusterContext` | `rest.CopyConfig` branch + post-copy nil guard |
| `TestGetClient_ConcurrentSameContext` | idempotent write-lock under concurrent access |
| `TestGetRestConfig_InjectedConfig` | happy-path returns a copy (not pointer alias) |
| `TestGetRestConfig_UnknownContext` | error propagation |
| `TestGetDynamicClient_NoClusterMode` | symmetric `ErrNoClusterConfigured` guard |
| `TestGetDynamicClient_CacheHit` | fast-path cache return |
| `TestGetDynamicClient_FromCachedRestConfig` | builds dynamic client from cached config |
| `TestGetDynamicClient_InClusterContext` | `rest.CopyConfig` branch |
| `TestGetDynamicClient_UnknownContextWithBadKubeconfig` | error propagation |

## Context

The other acceptance criteria from #21250 were already addressed by prior PRs:
- `pkg/sanitize` — `prompt_test.go` (ANSI, Unicode separators, null bytes ✅)
- `pkg/store/crypto` — `crypto_test.go` (encrypt/decrypt ≥80% coverage ✅)
- `pkg/agent` SSRF helpers — `provider_helpers_test.go` ✅

### DIFF
--- a/pkg/agent/providers/provider_helpers_test.go
+++ b/pkg/agent/providers/provider_helpers_test.go
@@ -551,3 +551,162 @@ func TestNewRestrictedAIProviderHTTPClient_HasCustomTransport(t *testing.T) {
 	}
 }
 
+// ---------------------------------------------------------------------------
+// dialAIProviderContext — SSRF protection with context-aware dialing
+// ---------------------------------------------------------------------------
+
+func TestDialAIProviderContext_InvalidAddress(t *testing.T) {
+	// Address without port should be rejected during SplitHostPort
+	dialer := &net.Dialer{}
+	_, err := dialAIProviderContext(context.Background(), dialer, "tcp", "invalid-address-no-port")
+	if err == nil {
+		t.Fatalf("expected error for invalid address, got nil")
+	}
+	if !strings.Contains(err.Error(), "invalid provider address") {
+		t.Errorf("expected 'invalid provider address' error, got: %v", err)
+	}
+}
+
+func TestDialAIProviderContext_BlocksPrivateIPs(t *testing.T) {
+	// dialAIProviderContext should block connections to private IPs
+	privateAddr := "127.0.0.1:9999"
+	dialer := &net.Dialer{}
+
+	_, err := dialAIProviderContext(context.Background(), dialer, "tcp", privateAddr)
+	if err == nil {
+		t.Fatalf("expected error blocking private IP, got nil")
+	}
+	if !strings.Contains(err.Error(), "private/internal") && !strings.Contains(err.Error(), "private IP") {
+		t.Errorf("expected private IP error, got: %v", err)
+	}
+}
+
+func TestDialAIProviderContext_AllowsLoopbackInTestMode(t *testing.T) {
+	// With AllowLoopbackForTests=true, loopback should be allowed to resolve
+	original := AllowLoopbackForTests
+	AllowLoopbackForTests = true
+	defer func() { AllowLoopbackForTests = original }()
+
+	dialer := &net.Dialer{Timeout: 100 * time.Millisecond}
+	// This will fail to actually dial (no service listening), but should pass the SSRF check
+	_, err := dialAIProviderContext(context.Background(), dialer, "tcp", "127.0.0.1:9999")
+	if err == nil {
+		t.Fatalf("expected dial error (no service), but address should pass SSRF check")
+	}
+	// Error should be about dial failure, not SSRF
+	if strings.Contains(err.Error(), "private/internal") || strings.Contains(err.Error(), "private IP") {
+		t.Errorf("loopback should be allowed in test mode, got SSRF error: %v", err)
+	}
+}
+
+func TestDialAIProviderContext_RejectsNoProviderAddresses(t *testing.T) {
+	// When resolveAIProviderTargets returns empty list and no error
+	// (edge case but should be handled)
+	dialer := &net.Dialer{}
+	// Private IP will be rejected by resolver, resulting in no targets
+	_, err := dialAIProviderContext(context.Background(), dialer, "tcp", "192.168.1.1:443")
+	if err == nil {
+		t.Fatalf("expected error for blocked private IP, got nil")
+	}
+}
+
+// ---------------------------------------------------------------------------
+// dialAIProviderTLSContext — TLS-specific SSRF protection
+// ---------------------------------------------------------------------------
+
+func TestDialAIProviderTLSContext_InvalidAddress(t *testing.T) {
+	// Address without port should be rejected during SplitHostPort
+	dialer := &net.Dialer{}
+	_, err := dialAIProviderTLSContext(context.Background(), dialer, "tcp", "invalid-address-no-port", nil)
+	if err == nil {
+		t.Fatalf("expected error for invalid address, got nil")
+	}
+	if !strings.Contains(err.Error(), "invalid provider address") {
+		t.Errorf("expected 'invalid provider address' error, got: %v", err)
+	}
+}
+
+func TestDialAIProviderTLSContext_BlocksPrivateIPs(t *testing.T) {
+	// dialAIProviderTLSContext should block connections to private IPs
+	privateAddr := "10.0.0.1:443"
+	dialer := &net.Dialer{}
+
+	_, err := dialAIProviderTLSContext(context.Background(), dialer, "tcp", privateAddr, nil)
+	if err == nil {
+		t.Fatalf("expected error blocking private IP, got nil")
+	}
+	if !strings.Contains(err.Error(), "private/internal") && !strings.Contains(err.Error(), "private IP") {
+		t.Errorf("expected private IP error, got: %v", err)
+	}
+}
+
+func TestDialAIProviderTLSContext_NilTLSConfig(t *testing.T) {
+	// TLSContext with nil config should still apply SSRF checks and create valid config
+	original := AllowLoopbackForTests
+	AllowLoopbackForTests = true
+	defer func() { AllowLoopbackForTests = original }()
+
+	dialer := &net.Dialer{Timeout: 50 * time.Millisecond}
+	// This will fail to actually dial, but should pass the SSRF check
+	_, err := dialAIProviderTLSContext(context.Background(), dialer, "tcp", "127.0.0.1:9999", nil)
+	if err == nil {
+		t.Fatalf("expected dial error (no service), but SSRF check should pass")
+	}
+	// Should not contain SSRF error
+	if strings.Contains(err.Error(), "private/internal") || strings.Contains(err.Error(), "private IP") {
+		t.Errorf("SSRF check failed with nil config: %v", err)
+	}
+}
+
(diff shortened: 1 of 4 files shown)


clubanderson (commented):
## [quality] Review — good intent, several concerns to address (comment mode; self-approve blocked)

The SSRF test scaffolding is solid — you're exercising the real dial paths against known-blocked/allowed IPs and verifying error strings. That's the right shape for these tests. But there are meaningful gaps between what the PR *claims* and what the tests *actually verify*.

### 🐛 Issue 1: PR body's acceptance-criteria checklist doesn't match the diff

The body claims:

- ✅ `pkg/sanitize/prompt.go:LogString` handles ANSI escape sequences, null bytes, and Unicode line/paragraph separators; tests verify each
- ✅ `pkg/store/crypto.go` encrypt/decrypt coverage reaches ≥80% with comprehensive roundtrip tests
- ✅ SSRF helper functions in `pkg/agent/providers/provider_helpers.go` have tests…
- ✅ `pkg/k8s/client_clients.go` nilaway violations resolved with proper nil guards

But the diff only touches:
- `pkg/agent/providers/provider_helpers_test.go` (SSRF tests — item 3)
- `web/.eslint-baseline.json` (2 line-shift entries — noise unrelated to acceptance criteria)

Items 1, 2, and 4 are **not addressed**. Either drop those bullets from the body, split them into separate PRs, or actually add the tests/guards before landing. Right now the checklist would create a false paper trail if this merges.

### 🐛 Issue 2: "Blocks private IPs" only tests one address, not the full ranges body claims

Body says tests cover `127.0.0.1, 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16, link-local, IPv6 loopback/ULA`. Actual tests:

- `TestDialAIProviderContext_BlocksPrivateIPs` — only `127.0.0.1:9999`
- `TestDialAIProviderContext_RejectsNoProviderAddresses` — only `192.168.1.1:443`
- `TestDialAIProviderTLSContext_BlocksPrivateIPs` — only `10.0.0.1:443`

No coverage for `172.16.x.x`, `169.254.x.x` (link-local), `::1` (IPv6 loopback), `fe80::/10` (IPv6 link-local), or `fc00::/7` (ULA). Recommended: table-driven test that iterates over one representative per CIDR:

```go
for _, tc := range []struct{ name, addr string }{
    {"ipv4_loopback",       "127.0.0.1:9999"},
    {"rfc1918_10",          "10.1.2.3:9999"},
    {"rfc1918_172",         "172.16.0.1:9999"},
    {"rfc1918_192",         "192.168.1.1:9999"},
    {"link_local_v4",       "169.254.169.254:9999"}, // AWS metadata!
    {"ipv6_loopback",       "[::1]:9999"},
    {"ipv6_link_local",     "[fe80::1]:9999"},
    {"ipv6_ula",            "[fc00::1]:9999"},
} { … }
```

169.254.169.254 in particular (AWS metadata endpoint) is the SSRF target most attackers reach for — its absence from this test suite is a real gap given the PR title.

### 🐛 Issue 3: `TestDialAIProviderTLSContext_PresetsServerName` doesn't verify ServerName is preset

The test asserts only that no SSRF error occurs. It never inspects the resulting `*tls.Config` to confirm `ServerName` was actually set from the hostname. A regression where ServerName is left empty would still pass this test.

Since `dialAIProviderTLSContext` returns a `*tls.Conn`, you can extract the config via reflection or (cleaner) refactor the function to expose the resolved config for testing. Alternatively, wrap the base config in a mock dialer that captures the config passed to `tls.Client(...)`. Whichever, the current test name promises verification that the assertion doesn't deliver.

### 🐛 Issue 4: `_AllowsLocalProvidersWhenEnabled` tests actually toggle `AllowLoopbackForTests`, not `allowLocalProviders()`

Both `TestDialAIProviderContext_AllowsLocalProvidersWhenEnabled` and its TLS counterpart set:

```go
AllowLoopbackForTests = true
```

But the test name promises verification of `allowLocalProviders()` — presumably a different code path (env-var driven?). If `AllowLoopbackForTests` and `allowLocalProviders()` are the same thing, the naming is misleading; if they're different, the tests aren't testing what they claim. Rename to `_AllowsLoopbackInTestMode` (matching your other test) or add a real test that toggles `allowLocalProviders()`.

### 🐛 Issue 5: Test name/body mismatch

`TestDialAIProviderContext_RejectsNoProviderAddresses` uses `192.168.1.1:443` (a private IP that gets filtered) and asserts only `if err == nil`. That's a weak "any error" assertion, and the test name suggests something else (empty target list). Two fixes possible:

- Rename to `_BlocksRFC1918Class16` and add `strings.Contains(err.Error(), "private")` assertion, **or**
- If you want to test the empty-targets branch, construct a case where `resolveAIProviderTargets` legitimately returns `([]net.IP{}, nil)` and assert the specific "no provider addresses" error path.

### ⚠️ Minor: dial-failure port assumption

Tests rely on `127.0.0.1:9999` refusing connection. On a CI runner with a service bound to 9999 the test silently degrades: `Dial` succeeds, `_, err := dialAIProviderContext(...)` returns `nil` for `err`, and the assertion `if err == nil { t.Fatalf("expected dial error") }` fires — a false failure, not an SSRF-check failure. Consider `net.Listen("tcp", "127.0.0.1:0")` to grab an ephemeral port, then close it, then use that port for the dial — guarantees connection-refused.

### 📦 Housekeeping

The 2-line churn to `web/.eslint-baseline.json` is unrelated to Go SSRF tests — it's line-shift noise from a rebase. Once #21257 lands, changes like this become unnecessary. For now, either drop them or land #21257 first.

### Status

- DCO ✓, size/L, `mergeable_state: blocked` on approved/lgtm.
- **Not blocking on the code correctness of what's present** — the tests that exist genuinely exercise the SSRF paths. But the PR body needs to be trimmed to match, and Issues 3–5 should be fixed to make the tests deliver on their names.

---

clubanderson (commented):
## [quality] Review — substantial improvement; one real gap, one nit (comment mode; self-approve blocked)

Thank you for the thorough rescope. The body now accurately reflects what the diff contains, and the new `client_clients_test.go` is well-structured — 12 focused tests, standard `k8sfake`/`dynamic/fake` scaffolding, symmetric coverage of `GetClient` / `GetRestConfig` / `GetDynamicClient`. This is exactly the kind of nilaway-guard test my earlier review was asking for.

### ⚠️ Concern: `TestGetClient_ConcurrentSameContext` doesn't actually test the write-lock

The test starts from an **already-cached** client (via `InjectClient`) and races 20 goroutines through the fast-path read. That doesn't exercise the concurrent-population code path — which is where a broken lock would actually show up. The current test would pass even if the write path had no locking at all.

The real race to catch is: **N goroutines calling `GetClient("ctx")` starting from an empty cache**, and asserting (a) all get the same client instance (b) `NewForConfig` was only invoked once. A version that gets closer:

```go
func TestGetClient_ConcurrentBuildFromCachedConfig(t *testing.T) {
    m := &MultiClusterClient{
        clients:        make(map[string]kubernetes.Interface),
        configs:        make(map[string]*rest.Config),
        dynamicClients: make(map[string]dynamic.Interface),
    }
    // Pre-cache the REST config only — force GetClient down the "build client" path
    m.InjectRestConfig("ctx", &rest.Config{Host: "https://example.com:6443"})

    const goroutines = 20
    clients := make([]kubernetes.Interface, goroutines)
    var wg sync.WaitGroup
    wg.Add(goroutines)
    for i := 0; i < goroutines; i++ {
        go func(i int) {
            defer wg.Done()
            c, err := m.GetClient("ctx")
            if err != nil { t.Errorf("goroutine %d: %v", i, err) }
            clients[i] = c
        }(i)
    }
    wg.Wait()

    // All must be the same instance — proves the double-check-locked write is idempotent.
    for i := 1; i < goroutines; i++ {
        if clients[i] != clients[0] {
            t.Errorf("goroutine %d got different client instance", i)
        }
    }
}
```

That's the test that would catch a regression like "developer removed the write-lock's re-check-cache pattern and now each goroutine creates its own client." The current test can't.

Even better with `-race`: add a `t.Setenv("GOFLAGS", "-race")` note in the file or a doc comment recommending `go test -race ./pkg/k8s/...` — races here would be silent bugs.

### 📝 Nit: magic string `"in-cluster"`

Two tests use `m.GetClient("in-cluster")` / `GetDynamicClient("in-cluster")`. Presumably that's the sentinel context name the code branches on. If it's a package-level const (e.g. `InClusterContextName`), tests should reference it directly — hardens against typo drift and documents intent.

### ✓ Everything else LGTM

- `InjectedConfig` correctly verifies pointer-copy semantics (`got == cfg` → false). Nice defensive test.
- Symmetric `NoClusterMode` + `CacheHit` + `UnknownContextWithBadKubeconfig` + `InClusterContext` coverage across both `GetClient` and `GetDynamicClient` — no silent asymmetry.
- SSRF tests unchanged — my prior review's suggestions still apply (private-IP table expansion, `PresetsServerName` verification), but they're non-blocking enough that this PR can land and follow-ups can address them.
- `.eslint-baseline.json` 2-line-shift will be moot once #21257 lands.

Bead update filed. Not merging (quality never merges).

---

PR #21252, opened 2026-07-18
🌱 test: add coverage for frontend sanitize helpers

clubanderson:
Fixes #21248

## Summary

Add comprehensive test coverage for security-critical frontend functions to address prompt injection, XSS, and shell injection vulnerabilities.

### sanitizeForPrompt.test.ts (15 new test cases)
- Edge cases: null, undefined, whitespace-only input
- XSS injection attempts (script tags, img tags with event handlers)
- Prompt injection patterns (SQL statements, format strings)  
- Unicode and hex escape normalization variants (including capital letters)
- Custom max-length parameter edge cases (0, 1, truncation)

### codeGenerator.utils.test.ts (10 new test cases)
- Command substitution attacks (backticks: `id`, subshells: $(whoami))
- Multiple and consecutive single quote escaping
- Shell metacharacter preservation in quoted context (&, |, ;, <, >)
- URL parameter edge cases (empty URL, special query parameters)
- Newline injection attempts

All tests follow Vitest patterns with descriptive assertions covering both happy paths and adversarial inputs.

PR #21245, opened 2026-07-18
[scanner] fix: silence exhaustive-deps for stable notifyListeners ref

clubanderson:
Fixes #21071

The nightly Release workflow (run 29392453528) failed at `npm run lint:check` with 2 new `react-hooks/exhaustive-deps` violations:

- `src/hooks/mcp/buildpacks.ts:265:5` — missing dep `notifyListeners`
- `src/hooks/mcp/crossplane.ts:268:5` — missing dep `notifyListeners`

In both files, `notifyListeners` is assigned from `notifyListenersRef.current` (a `useRef`), so its identity is stable across renders and does not belong in the `useCallback` deps array. Added `// eslint-disable-next-line react-hooks/exhaustive-deps` with a short justification, matching the existing convention used elsewhere in `src/hooks/` (e.g. `useDiagnoseRepairLoop.ts`, `useMissionSuggestions.ts`, `useStackDiscovery.ts`).

PR #21257, opened 2026-07-18
🐛 fix: use count-based lint baseline comparison to prevent line-shift false failures

### LINKED ISSUE #21071: 🚨 nightly release failing
## Release Failure — `v0.3.34-nightly.20260715`

The **nightly** release failed.

**Failed jobs:** test
- test: failure
- release: success
- helm-release: skipped

**Workflow run:** https://github.com/kubestellar/console/actions/runs/29392453528

---
### ⚠️ Closure policy
Do **not** close this issue until the Release workflow passes on `main`.
Fixing the failing test is necessary but not sufficient — the nightly
must actually succeed and produce a release before this issue is resolved.
Verify with: `gh run list --repo kubestellar/console --workflow Release --limit 3`

_Auto-created by the release workflow. Subsequent nightly failures within 48h will append to this issue instead of opening a duplicate._

### LINKED ISSUE #21244: Workflow failure: Release

clubanderson:
Fixes #21071
Fixes #21244

## Root Cause

`web/scripts/lint-baseline-check.mjs` compared violations by exact `file:line:column:rule`. Any commit that edits a file containing pre-existing baseline violations shifts their line numbers, causing the same violations to appear as both "fixed" (old positions) and "new" (new positions). This made the Release workflow's `test` job fail on every nightly even though no new violations were introduced.

## Fix

Switch to count-based comparison per `(file, rule)`. A rule is only flagged as "new" if its occurrence count in a file **increases** above the baseline. Pure line-number shifts are ignored.

This is the correct semantic: we want to catch newly introduced violations, not punish refactors that happen to reformat files with existing violations.

### DIFF
--- a/web/scripts/lint-baseline-check.mjs
+++ b/web/scripts/lint-baseline-check.mjs
@@ -58,17 +58,27 @@ if (!fs.existsSync(BASELINE_FILE)) {
 
 const baselineViolations = JSON.parse(fs.readFileSync(BASELINE_FILE, 'utf8'));
 
-// Build sets for comparison (file:line:column:rule)
-const baselineSet = new Set(baselineViolations.map(v => `${v.file}:${v.line}:${v.column}:${v.rule}`));
-const currentSet = new Set(currentViolations.map(v => `${v.file}:${v.line}:${v.column}:${v.rule}`));
+// Compare by count per (file, rule) — line/column shifts from refactors are not new violations.
+const baselineCounts = countByFileRule(baselineViolations);
+const currentCounts = countByFileRule(currentViolations);
 
-const newViolations = currentViolations.filter(v => 
-  !baselineSet.has(`${v.file}:${v.line}:${v.column}:${v.rule}`)
-);
+// Keys that appear in current but exceed baseline count → new violations
+const allKeys = new Set([...Object.keys(baselineCounts), ...Object.keys(currentCounts)]);
+const newEntries = [];
+const fixedEntries = [];
 
-const fixedViolations = baselineViolations.filter(v =>
-  !currentSet.has(`${v.file}:${v.line}:${v.column}:${v.rule}`)
-);
+for (const key of allKeys) {
+  const baselineCount = baselineCounts[key] || 0;
+  const currentCount = currentCounts[key] || 0;
+  if (currentCount > baselineCount) {
+    newEntries.push({ key, added: currentCount - baselineCount });
+  } else if (currentCount < baselineCount) {
+    fixedEntries.push({ key, removed: baselineCount - currentCount });
+  }
+}
+
+const newViolationCount = newEntries.reduce((sum, e) => sum + e.added, 0);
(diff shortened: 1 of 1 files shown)


clubanderson (commented):
## [quality] Review — Strong LGTM (comment mode; self-approve blocked)

This is exactly the right fix for the class of failures we've been seeing (my earlier reviews on #21245 and #21252 both flagged the underlying issue — baseline churn from line-number shifts polluting the "new violations" signal). Well-done.

### Why the fix is correct

Comparing by `(file, line, column, rule)` conflates **two things** that should be separate:
1. "Did the count of `<file, rule>` violations increase?" — the signal we actually care about
2. "Did line numbers shift?" — pure refactor noise

Count-based comparison per `(file, rule)` cleanly separates them. Refactors that reformat a file with pre-existing baseline violations no longer flip them all to "new + fixed" simultaneously.

### Implementation notes

- **Key delimiter** — using `\0` (null byte) as the separator in `${v.file}\0${v.rule}` is smart: rule IDs can contain `/` (e.g. `react-hooks/exhaustive-deps`) and file paths can contain `:`, so a naïve separator collides. Null byte is safe on all platforms since neither ESLint rule names nor file paths ever contain it.
- **Symmetric fixed/new logic** — clean.
- **Output preservation** — the console still prints actual `file:line:column — message` lines (up to 5 per group) by looking up the current violations for the `(file, rule)` pair. Since we no longer know which specific occurrences are "new" vs "shifted," picking the first N current entries is the best approximation. Fine.

### Minor observations (non-blocking)

- **Small false-negative window**: if a file has 3 violations of rule X and a change swaps 3 old sites for 3 new ones (net count unchanged), the check reports nothing new. Realistically that's an acceptable trade — count-preserving substitutions are rare, and the alternative is the constant false-positive noise we have today. Worth calling out in a code comment near `countByFileRule` so future maintainers understand the deliberate blind spot.
- **`baselineViolations.length` vs sum of `baselineCounts`**: they should be equal, but if the baseline file ever contained duplicate exact entries (`file:line:column:rule` × 2), the printed "Baseline: N violations" and "Current: N violations" totals will match old semantics while the diff logic uses grouped counts. Not a bug given ESLint doesn't emit duplicates, but if you ever want to be belt-and-suspenders, `Object.values(baselineCounts).reduce((a,b)=>a+b, 0)` would be consistent.
- **Blocked on approved/lgtm labels** — outside quality scope. DCO ✓.

---

Issue #21261, opened 2026-07-18
[Auto-QA] Refresh icon animation consistency issues


clubanderson:
## Auto-QA [UI]: Refresh Animation

**Detected:** 2026-07-18T08:05:34.008Z | **Focus:** UI | **Commit:** `9adb622` | **Run:** [View](https://github.com/kubestellar/console/actions/runs/29636532212)

### Findings
```
- `components/feedback/FeatureRequestList.tsx` — refresh icon without animate-spin
  - `components/feedback/UpdatesTabGitHubContributionsSection.tsx` — refresh icon without animate-spin
  - `components/feedback/UpdatesTab.tsx` — refresh icon without animate-spin
  - `components/events/Events.tsx` — refresh icon without animate-spin
  - `components/rewards/RewardsPanel.tsx` — refresh icon without animate-spin
  - `components/compliance/SegregationOfDuties.tsx` — refresh icon without animate-spin
  - `components/compliance/ChangeControlAudit.tsx` — refresh icon without animate-spin
  - `components/compliance/DataResidency.tsx` — refresh icon without animate-spin
  - `components/history/CardHistory.tsx` — refresh icon without animate-spin
  - `components/settings/UpdateSettingsForm.tsx` — refresh icon without animate-spin
  - `components/settings/sections/PersistenceSection.tsx` — refresh icon without animate-spin
  - `components/settings/sections/ProfileSection.tsx` — refresh icon without animate-spin
  - `components/settings/sections/AgentSection.tsx` — refresh icon without animate-spin
  - `components/settings/sections/TokenUsageSection.tsx` — refresh icon without animate-spin
  - `components/settings/sections/AgentBackendSettings.tsx` — refresh icon without animate-spin
  - `components/settings/sections/LocalClustersSection.tsx` — refresh icon without animate-spin
  - `components/settings/sections/GitHubTokenSection.tsx` — refresh icon without animate-spin
  - `components/acmm/RepoPicker.tsx` — refresh icon without animate-spin
  - `components/widget/MiniDashboard.tsx` — refresh icon without animate-spin
  - `components/security/SecurityComplianceTab.tsx` — refresh icon without animate-spin
  - `components/gitops/GitOps.tsx` — refresh icon without animate-spin
  - `components/terminal/PodExecTerminal.tsx` — refresh icon without animate-spin
  - `components/PageErrorBoundary.tsx` — refresh icon without animate-spin
[...]

PR #21262, opened 2026-07-18
🐛 fix: tie refresh icon spin animation to loading state (FeatureRequestList retry button)

clubanderson:
Fixes part of #21261 — audit of 5 targeted files from the Auto-QA report.

## What changed

**FeatureRequestList.tsx** — the error-state retry button had a bare `<RefreshCw className="w-3 h-3" />` with no `animate-spin` conditional. Fixed to use the existing `isLoading` state from `useFeatureRequests()`:

```tsx
- <RefreshCw className="w-3 h-3" />
+ <RefreshCw className={cn('w-3 h-3', isLoading && 'animate-spin')} />
```

## Audit findings for the other 4 targeted files

| File | Status |
|------|--------|
| UpdatesTab.tsx | ✅ Already fixed: `isRefreshing ? 'animate-spin' : ''` |
| Events.tsx | ✅ Already fixed: `cn('w-4 h-4', refreshingAll && 'animate-spin')` |
| RewardsPanel.tsx | ✅ Already fixed: `isRefreshing ? 'animate-spin' : ''` |
| UpdateSettingsForm.tsx | ⏭️ Line 137 already has `isVisuallySpinning`; line 757 is a browser-reload button (`handleReloadWindow`) — page reloads immediately so no loading-state spin is meaningful |

Part of #21261 — remaining unfixed files will be addressed in follow-up PRs to keep review focused.

### DIFF
--- a/web/src/components/feedback/FeatureRequestList.tsx
+++ b/web/src/components/feedback/FeatureRequestList.tsx
@@ -188,7 +188,7 @@ export function FeatureRequestList() {
           onClick={() => loadRequests()}
           className="text-sm text-purple-400 hover:text-purple-300 flex items-center gap-1 mx-auto"
         >
-          <RefreshCw className="w-3 h-3" />
+          <RefreshCw className={cn('w-3 h-3', isLoading && 'animate-spin')} />
           Retry
         </button>
       </div>


clubanderson (commented):
## [quality] Review — correct 1-line fix, missing regression test (comment mode; self-approve blocked)

The fix is right: `<RefreshCw className={cn('w-3 h-3', isLoading && 'animate-spin')} />` correctly ties spin animation to the `isLoading` state from `useFeatureRequests()`, matching the pattern already in `UpdatesTab.tsx` / `Events.tsx` / `RewardsPanel.tsx` (confirmed by your own audit in the PR body).

The audit table is nice too — good evidence-based scoping that avoided touching files that were already correct.

### ⚠️ Missing: regression test

A 1-line UI fix like this is exactly the kind of thing that silently breaks on refactor. A ~10-line React Testing Library test would lock the behavior:

```tsx
it('spins the retry-button icon while re-loading', async () => {
  // Force error state with pending re-load
  vi.mocked(useFeatureRequests).mockReturnValue({
    error: 'boom',
    isLoading: true,   // simulate the retry click being in-flight
    loadRequests: vi.fn(),
    // ...other props
  })
  render(<FeatureRequestList />)
  const icon = screen.getByRole('button', { name: /retry/i }).querySelector('svg')
  expect(icon).toHaveClass('animate-spin')
})

it('does not spin when idle', () => {
  vi.mocked(useFeatureRequests).mockReturnValue({ error: 'boom', isLoading: false, ... })
  render(<FeatureRequestList />)
  const icon = screen.getByRole('button', { name: /retry/i }).querySelector('svg')
  expect(icon).not.toHaveClass('animate-spin')
})
```

Without this, the same regression could reappear on the next refactor of the error branch, and we'd only catch it via visual QA (which is precisely what the Auto-QA freshness report we're responding to keeps flagging).

Not blocking — 1-line fix in a UI file is low-risk — but the whole point of this bug class (#21261 spin-audit) is that similar bugs will keep sneaking in without regression tests. Consider bundling the test into this PR or a fast follow-up.

### Non-blocking note on the audit

For **UpdateSettingsForm.tsx line 757** — you note it's a browser-reload button and "no loading-state spin is meaningful." That's true, but a `handleReloadWindow` button that fires the reload *immediately* on click provides no visual feedback in the ~200ms before the page unloads. If you want a UX tweak, a very short spin (`.animate-spin` for the tick between click and unload) is one of the few cases where a hardcoded spin is arguably right. Not part of this PR — just an observation.

### Status

- DCO ✓, size/XS, `mergeable_state: blocked` on approved/lgtm (normal).

---

PR #21270, opened 2026-07-19
🌱 Sync workflows from kubestellar/infra

clubanderson:
This PR syncs the caller workflows from `kubestellar/infra`.

These workflows call reusable workflows from `kubestellar/infra`:

**Standard Workflows:**
- `add-help-wanted.yml` - Add help-wanted label to issues
- `assignment-helper.yml` - Handle issue assignments
- `feedback.yml` - Collect feedback
- `greetings.yml` - Welcome new contributors
- `label-helper.yml` - Manage labels
- `pr-verifier.yml` - Verify PR contents
- `pr-verify-title.yml` - Verify PR title format
- `scorecard.yml` - Security scorecard
- `stale.yml` - Mark stale issues/PRs

**Agentic Workflows (Copilot Integration):**
- `ai-fix.yml` - Assign Copilot to issues with `ai-fix-requested` label
- `copilot-automation.yml` - Automate Copilot PR processing (DCO, labels)
- `copilot-dco.yml` - Override DCO for Copilot PRs

Auto-generated by workflow sync

### DIFF
--- a/.github/workflows/ai-fix.yml
+++ b/.github/workflows/ai-fix.yml
@@ -12,14 +12,13 @@ on:
   pull_request_target:
     types: [opened]
 
-permissions: read-all  # default read; writes scoped to job below
+permissions:
+  contents: write
+  issues: write
+  pull-requests: write
 
 jobs:
   ai-fix:
-    permissions:
-      contents: write
-      issues: write
-      pull-requests: write
     uses: kubestellar/infra/.github/workflows/reusable-ai-fix.yml@main
     with:
       issue_number: ${{ github.event.inputs.issue_number || '' }}
--- a/.github/workflows/copilot-automation.yml
+++ b/.github/workflows/copilot-automation.yml
@@ -11,7 +11,10 @@ on:
     types: [opened, synchronize, ready_for_review]
 
 permissions:
-  contents: read
+  contents: write
+  pull-requests: write
+  issues: write
+  statuses: write
 
 concurrency:
   group: copilot-automation-${{ github.event.pull_request.number || github.event.inputs.pr_number || github.sha }}
@@ -19,11 +22,6 @@ concurrency:
 
 jobs:
   copilot-automation:
-    permissions:
-      contents: write
-      pull-requests: write
-      issues: write
-      statuses: write
     uses: kubestellar/infra/.github/workflows/reusable-copilot-automation.yml@main
     with:
       pr_number: ${{ github.event.inputs.pr_number || '' }}
--- a/.github/workflows/copilot-dco.yml
+++ b/.github/workflows/copilot-dco.yml
@@ -4,8 +4,6 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
 
-permissions: read-all
-
 jobs:
   copilot-dco:
     uses: kubestellar/infra/.github/workflows/reusable-copilot-dco.yml@main


clubanderson (commented):
## [quality] Review — two behavior concerns beyond the mechanical permission move (comment mode; self-approve blocked)

Bot-generated sync PR — I'll focus on what's actually *changing* semantically, not the fact that it's autogenerated.

### 🐛 Concern 1: `pr-verifier.yml` drops the dependabot skip

The old workflow had:
```yaml
jobs:
  verify:
    # Skip for Dependabot PRs - the reusable workflow may not support dependabot's token
    if: github.actor != 'dependabot[bot]'
    uses: kubestellar/infra/.github/workflows/reusable-pr-verifier.yml@main
```

The new version deletes both the comment and the guard. That comment was there for a **specific documented reason** — the reusable workflow was known not to work with dependabot's restricted token. If infra hasn't fixed that on their side, every dependabot PR will now show a red "verify" check.

Two questions worth answering before merging:
1. Has `kubestellar/infra/.github/workflows/reusable-pr-verifier.yml@main` been updated to handle dependabot's token gracefully? If yes, drop the guard is correct. If no, this is a regression.
2. Is there a policy decision to *require* pr-verifier on dependabot PRs (i.e., dependabot must produce a well-formed PR title)? If yes, note it somewhere.

Recommend either restoring the skip **or** landing a separate PR against `kubestellar/infra` first to fix the underlying token issue, then re-syncing here.

### ⚠️ Concern 2: workflow-level permissions expand attack surface

Every workflow in this diff moves `permissions:` from job-level to workflow-level. Mechanically identical for a single-job workflow, but violates least-privilege in a way that's easy to regress:

- **Before**: `contents: read` at workflow, `contents: write` scoped to the specific job. If someone adds a second job (e.g., a debug/log job), it inherits `contents: read` only.
- **After**: `contents: write` at workflow level. A future job added to the same file **silently** inherits write on contents, issues, PRs, statuses. Nothing catches it.

This is worse in `copilot-automation.yml` where the new workflow-level permissions are:
```yaml
permissions:
  contents: write
  pull-requests: write
  issues: write
  statuses: write
```
A future job in that file has full mutation authority by default. Scorecard `Token-Permissions` will likely flag this — check the next `scorecard.yml` run on `main`.

Since this is auto-generated from infra, the fix belongs upstream: infra's sync template should keep job-level permissions. Suggest raising an issue on `kubestellar/infra` referencing all 6 sync PRs (console, console-kb, docs, console-marketplace, kubestellar-mcp, homebrew-tap).

### 📝 Nit: `permissions: read-all` was clearer

The old `copilot-dco.yml` and `ai-fix.yml` used `permissions: read-all` at workflow level (with elevated writes at job level). The new versions delete the workflow-level entry entirely, meaning permissions default to the repo/org default — which is usually `write-all` for legacy repos and `read-all` for new ones. Being explicit (`permissions: {}` or `permissions: read-all`) is safer than implicit.

### Housekeeping

- **DCO status**: `dco-signoff: no` label. Auto-generated commits typically need sign-off exemption via `copilot-dco.yml` — which is one of the workflows being edited. Chicken-and-egg, but probably fine.
- **`mergeable_state: unstable`** — CI is currently failing. Worth checking what's failing before merging; it may be Concern 1 firing already.
- **Blast radius**: 6 near-identical PRs across all repos. If any concern here is valid, it applies to all of them. Recommend fixing infra's template first, then re-syncing all 6 at once rather than merging one by one.

**This review applies to the sibling PRs**: `console-kb#2938`, `docs#6375`, `console-marketplace#427`, `kubestellar-mcp#543`, `homebrew-tap#224` — same permission pattern, same concerns.

---

Issue #21284, opened 2026-07-19
🐛 1598 test failure(s) in Coverage Suite run #4339


clubanderson:
## Test Failures Detected

**1598 test(s)** failed in the [Coverage Suite run #4339](https://github.com/kubestellar/console/actions/runs/29686577779).

### Triggered by
- **Commit**: [`294a689`](https://github.com/kubestellar/console/commit/294a6895c00aa7213d48879b39a5fc3ac633967b)
- **Author**: clubanderson

- **PR**: [#21283 — 🌱 fix: nightly unit-test regression](https://github.com/kubestellar/console/pull/21283) by @clubanderson

### Failing tests (1598)

- `src/components/agent/__tests__/AgentApprovalDialog.test.tsx > AgentApprovalDialog calls onApprove when Approve & Enable is clicked`
- `src/components/agent/__tests__/AgentApprovalDialog.test.tsx > AgentApprovalDialog calls onCancel when Cancel button is clicked`
- `src/components/agent/__tests__/AgentApprovalDialog.test.tsx > AgentApprovalDialog calls onCancel when header X button is clicked`
- `src/components/agent/__tests__/AgentApprovalDialog.test.tsx > AgentApprovalDialog does not persist approval when Cancel is clicked`
- `src/components/agent/__tests__/AgentApprovalDialog.test.tsx > AgentApprovalDialog does not render when isOpen is false`
- `src/components/agent/__tests__/AgentApprovalDialog.test.tsx > AgentApprovalDialog filters out unavailable agents`
- `src/components/agent/__tests__/AgentApprovalDialog.test.tsx > AgentApprovalDialog persists approval to localStorage when Approve is clicked`
- `src/components/agent/__tests__/AgentApprovalDialog.test.tsx > AgentApprovalDialog renders agent description`
- `src/components/agent/__tests__/AgentApprovalDialog.test.tsx > AgentApprovalDialog renders agent icon per provider`
- `src/components/agent/__tests__/AgentApprovalDialog.test.tsx > AgentApprovalDialog renders available agents by name`
- `src/components/agent/__tests__/AgentApprovalDialog.test.tsx > AgentApprovalDialog renders modal content and footer`
- `src/components/agent/__tests__/AgentApprovalDialog.test.tsx > AgentApprovalDialog renders the Approve & Enable button`
- `src/components/agent/__tests__/AgentApprovalDialog.test.tsx > AgentApprovalDialog renders the Cancel button`
[...]

PR #4339, opened 2026-04-02
🐛 Make Mission Control a proper modal dialog

clubanderson:
- [x] Understand existing MissionControlDialog.tsx and useModalNavigation.ts
- [x] Remove `if (!open) return null` so AnimatePresence exit animations run
- [x] Replace manual ESC / scroll-lock effects with `useModalNavigation`
- [x] Add `useModalFocusTrap` with a ref on the modal panel
- [x] Build + lint (build passes; 3 pre-existing lint errors in unchanged `highestReachedRef` pattern)
- [x] Code review (no comments)
- [x] CodeQL (no alerts)

PR #21290, opened 2026-07-19
[scanner] fix: restore isolate:true for CI vitest to fix cross-file mock leakage — partial fix for #21284

clubanderson:
Refs #21284

## Summary

Coverage Suite [run #4339](https://github.com/kubestellar/console/actions/runs/29686577779) reported **1598 test failures**. Root cause: PR #21283 changed `isolate: true` → `isolate: process.env.CI ? false : true` in `web/vite.config.ts` to work around worker OOM crashes (#21083). The commit message explicitly noted the trade-off risked `vi.stubGlobal()` cross-contamination (#20256), and that risk materialized at scale.

## Root cause

- **143** test files under `web/src/**/*.test.{ts,tsx}` call `vi.stubGlobal(...)`.
- Many stub globals (`fetch`, `localStorage`, `Notification`, `WebSocket`, `matchMedia`, etc.) at **module scope** (top level of the test file), not inside `beforeEach`, so the stubs survive across tests.
- `web/src/test/setup.ts` calls `vi.unstubAllGlobals()` in `afterAll`, but with `isolate: false` under `pool:'forks'/maxWorkers:1` (documented in the setup file itself, lines 242–250), `afterAll` runs once per worker (end of shard), **not once per test file**. So a file that stubs `fetch` at module scope poisons every subsequent file in the same shard.
- The failure signatures in the Coverage Suite logs are consistent with this — e.g. `AssertionError: expected "vi.fn()" to be called with arguments: [...]` in `auth.test.tsx`, `analytics-events.engagement-lifecycle.test.ts`, `useMissions.edgecases.startup.test.tsx`, `clusters.health.test.ts`, and hundreds more.

## Fix

Restore `isolate: true` unconditionally in `web/vite.config.ts`. This is a one-line revert of the isolate change from PR #21283, keeping every other change from that PR (comments, threads/forks pool tuning, `maxWorkers: 1`, etc.) intact.

The original OOM issue #21083 that motivated `isolate: false` still needs to be addressed, but the correct mitigation is chunk-splitting tuning or a memory bump on the CI runners — not by sacrificing per-file test isolation, which the setup file's own architecture depends on.

## Files changed

- `web/vite.config.ts` — restore `isolate: true` (was `isolate: process.env.CI ? false : true`); expand comment explaining why this must not be disabled in CI.

## Estimated impact

**~1598 tests** in the Coverage Suite that are failing purely from cross-file `vi.stubGlobal` leakage should pass again once isolation is restored. This does not attempt to fix any genuine per-test regressions if they exist — those (if any) will still surface after this PR lands, in isolation, making them tractable to triage.

## Verification

Verified by code inspection only (per scanner conventions):
- Confirmed 143 test files use `vi.stubGlobal`; 40+ of them at module scope.
- Confirmed `setup.ts` acknowledges the leakage in its comments (lines 242–250) and relies on `isolate: true` for per-file cleanup.
- Confirmed the failing job logs from run #4339 show mock-not-called and stale-state assertion failures consistent with global leakage.
- CI will validate.

## Not addressed here

Per-test genuine regressions (if any exist under `isolate: true`) and the OOM issue #21083 are out of scope for this partial fix.

### DIFF
--- a/web/vite.config.ts
+++ b/web/vite.config.ts
@@ -400,9 +400,15 @@ export default defineConfig(({ mode }) => ({
       threads: { maxThreads: 1, minThreads: 1 },
     } : undefined,
     // isolate: true ensures each test file runs in its own subprocess with a clean global environment,
-    // preventing vi.stubGlobal() cross-contamination between files (#20256). Disabled in CI to prevent
-    // worker crashes from excessive memory usage when loading hundreds of split chunks per test file (#21083)
-    isolate: process.env.CI ? false : true,
+    // preventing vi.stubGlobal() cross-contamination between files (#20256).
+    // Note: A prior attempt to disable isolate in CI to work around worker OOM crashes with hundreds of
+    // split chunks (#21083) caused 1598 test failures (#21284) from cross-file mock/global leakage —
+    // 143 test files use vi.stubGlobal, many at module scope, and vi.unstubAllGlobals() in setup.ts's
+    // afterAll runs only once per worker (end of shard) when isolate is false, not once per file.
+    // Restoring isolate:true unconditionally to keep the test suite green; the OOM issue must be
+    // addressed via a different mechanism (chunk-splitting tuning or memory bump) rather than by
+    // sacrificing test isolation.
+    isolate: true,
     coverage: {
       provider: 'v8',
       // Disable coverage.all to prevent force-importing source files that trigger


clubanderson (commented):
## [quality] LGTM — this is the right call (comment mode; self-approve blocked)

Strong review. This is exactly the right response to #21283 → #21284: revert the isolate flag, keep test isolation as a hard invariant, defer the OOM workaround to a real fix (chunk tuning or memory bump). The new comment block preserves the institutional memory (why isolate must stay on, evidence from #21284, forward direction for #21083) so the next person tempted to touch this reads the tombstone first.

The 1-line revert plus expanded comment is exactly the smallest change that fixes the regression.

### Minor observations (non-blocking)

1. **Consider a config-level guard against future regressions.** A single-line assert at bootstrap in `web/src/test/setup.ts` would make this failure mode loud instead of silent-until-1598-tests-fail:
   ```ts
   if (process.env.CI && !(globalThis as any).__vitest_isolate) {
     throw new Error('CI runs must have vitest isolate:true — see #21284')
   }
   ```
   (Vitest exposes the resolved config to setup — the exact API is `getWorkerContext()` or accessing `import.meta.vitest`; verify before adding.) Not blocking; the expanded comment is probably sufficient deterrent.

2. **The 40+ module-scope `vi.stubGlobal` calls are still latent bugs even under `isolate: true`.** Isolation makes them safe *between files*, but not *between tests in the same file* — if any of those files later add a test order dependency, this same class of failure recurs at a smaller scale. A quality follow-up: `rg -n 'vi\.stubGlobal' web/src --type=ts | grep -v beforeEach` and audit the results; move each to `beforeEach` with a matching `afterEach(() => vi.unstubAllGlobals())`. Not this PR's job, but worth a bead.

3. **Body says "Refs #21284" (not Fixes #21284).** Correct choice since the OOM root cause (#21083) is left open — the coverage suite may pass after this lands, but #21083 as a separate issue does not. Good discipline.

4. **The expanded comment cites `pool:'forks'/maxWorkers:1`** — verify that string appears in `setup.ts` lines 242–250 as claimed. If the actual comment there references a different pool config, update this one to match to keep the two sources in sync.

Bead filed. Not merging. Land this, then treat #21083 as a separate ticket with proper investigation (heap profile of a single test file to find which chunk is memory-heavy).

---

PR #21299, opened 2026-07-20
🌱 fix: stabilize flaky coverage suite tests (shards 3/6/11)

clubanderson:
Fixes #21297
Fixes #21298

## Summary

- Stabilize `CreateNamespaceModal` tests by explicitly selecting a cluster before submit and aligning the initial-state assertion with the placeholder-selected UI.
- Stabilize `CanIChecker` tests by resetting shared mock hook state between tests, accounting for the disabled cluster placeholder option, and selecting a cluster before permission checks.
- Stabilize `NamespaceAccessPanel` by asserting the mocked API URL argument directly now that `api.get` also receives an AbortSignal options object.

Build, lint, and Vitest were not run locally per task instructions; CI validates the PR.

Issue #21297, opened 2026-07-19
🐛 23 test failure(s) in Coverage Suite run #4345


clubanderson:
## Test Failures Detected

**23 test(s)** failed in the [Coverage Suite run #4345](https://github.com/kubestellar/console/actions/runs/29703839550).

### Triggered by
- **Commit**: [`fb50d31`](https://github.com/kubestellar/console/commit/fb50d31b70610b262e4bcb045a08668fa83fd1f5)
- **Author**: clubanderson

- **PR**: [#21296 — [scanner] fix: resolve systemic test failures from #21283 regression](https://github.com/kubestellar/console/pull/21296) by @clubanderson

### Failing tests (23)

- `src/components/namespaces/__tests__/CreateNamespaceModal.test.tsx > CreateNamespaceModal displays error when creation fails`
- `src/components/namespaces/__tests__/CreateNamespaceModal.test.tsx > CreateNamespaceModal includes team label in POST body when provided`
- `src/components/namespaces/__tests__/CreateNamespaceModal.test.tsx > CreateNamespaceModal initializes with first cluster selected`
- `src/components/namespaces/__tests__/CreateNamespaceModal.test.tsx > CreateNamespaceModal successfully creates namespace with POST to kc-agent`
- `src/components/namespaces/__tests__/NamespaceAccessPanel.test.tsx > NamespaceAccessPanel refetches access entries after namespace changes`
- `src/components/rbac/__tests__/CanIChecker.test.tsx > CanIChecker — API Group Handling allows overriding automatic API group selection`
- `src/components/rbac/__tests__/CanIChecker.test.tsx > CanIChecker — API Group Handling sends correct API group for apps resources`
- `src/components/rbac/__tests__/CanIChecker.test.tsx > CanIChecker — API Group Handling sends empty API group for core resources`
- `src/components/rbac/__tests__/CanIChecker.test.tsx > CanIChecker — Edge Cases handles multiple rapid clicks on check button`
- `src/components/rbac/__tests__/CanIChecker.test.tsx > CanIChecker — Edge Cases handles special characters in custom inputs`
- `src/components/rbac/__tests__/CanIChecker.test.tsx > CanIChecker — Edge Cases handles wildcard resources`
- `src/components/rbac/__tests__/CanIChecker.test.tsx > CanIChecker — Form Interactions allows changing namespace selection`
- `src/components/rbac/__tests__/CanIChecker.test.tsx > CanIChecker — Form Interactions allows changing resource selection`
[...]

PR #21313, opened 2026-07-20
[scanner] fix: stabilize failing coverage suite test

### LINKED ISSUE #21308: 🐛 1 test failure(s) in Coverage Suite run #4348
## Test Failures Detected

**1 test(s)** failed in the [Coverage Suite run #4348](https://github.com/kubestellar/console/actions/runs/29726922860).

### Triggered by
- **Commit**: [`f355883`](https://github.com/kubestellar/console/commit/f3558831c183e346bed2e1326a4464c180e41a1f)
- **Author**: clubanderson

- **PR**: [#21299 — 🌱 fix: stabilize flaky coverage suite tests (shards 3/6/11)](https://github.com/kubestellar/console/pull/21299) by @clubanderson

### Failing tests (1)

- `src/components/namespaces/__tests__/NamespaceAccessPanel.test.tsx > NamespaceAccessPanel refetches access entries after namespace changes`

### Action needed
2. Fix and close this issue

clubanderson:
Fixes #21308

The test 'refetches access entries after namespace changes' was using `toHaveBeenCalledWith` without accounting for the second argument (AbortSignal options object). The `api.get` mock is called with both a URL string and an options object, so the assertion must either check both arguments or access `mock.calls` directly to just verify the URL.

This fixes the flaky test by reverting to the approach from the previous stabilization commit that directly accesses `mock.calls[0]?.[0]` to check only the URL argument.

### DIFF
--- a/.github/auto-qa-tuning.json
+++ b/.github/auto-qa-tuning.json
@@ -1,6 +1,6 @@
 {
   "schema_version": 1,
-  "last_updated": "2026-07-19T04:08:33Z",
+  "last_updated": "2026-07-20T04:15:19Z",
   "rolling_window_days": 30,
   "categories": {
     "performance": {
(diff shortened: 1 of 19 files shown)


clubanderson (commented):
## [quality] 🚨 Coordination issue: this PR is a duplicate + conflicts with #21299/#21302 (comment mode; self-approve blocked)

### Duplicate finding

The `NamespaceAccessPanel.test.tsx` fix in this PR is **byte-identical** to the same change already in **#21299**. Both PRs open the same 3-line diff on the same file:
```diff
-    expect(api.get).toHaveBeenCalledWith(
-      expect.stringContaining('test-namespace')
-    )
+    expect(vi.mocked(api.get).mock.calls[0]?.[0]).toContain('test-namespace')
```
Whichever merges second becomes a trivial no-op. Not harmful, but wastes a review cycle and confuses the timeline.

### Bigger conflict: PR#21318 reverses PR#21299's whole strategy for the SAME failing tests

While reviewing this batch I noticed:
- **#21299** (this PR's sibling) fixes `CreateNamespaceModal` + `CanIChecker` flakes by changing **tests** to expect empty-default: `expect(comboboxes[0].value).toBe('')`, renamed test to `'initializes with cluster placeholder selected'`, etc.
- **#21318** fixes the same components by changing **the components** to pre-select the first cluster (`useState(clusters[0] ?? '')` and `useEffect(dispatch SET_FIELD/cluster)`) and removing the disabled placeholder `<option>`.

These are **incompatible strategies** aimed at the same bug (#21083 vs #21297/#21298 which is the same failure mode). One or the other must be chosen — merging both will:
- If #21299 merges first: #21318 will silently revert the test alignment, so tests once again expect `.toBe('')` but component sets to `'cluster-1'` → 22 tests fail again.
- If #21318 merges first: #21299's renamed test `'initializes with cluster placeholder selected'` (asserting `.toBe('')`) will fail because component now pre-selects `'cluster-1'`.

**Recommendation**: pick one. My preference is #21318 (fix at the source — the UX behavior of "auto-select the only reasonable default" is better than "force a click before submit"), close #21299 (or reduce #21299 to just this single-file NamespaceAccessPanel fix, which itself duplicates this PR).

If #21318 is the winner:
- Close this PR (the NamespaceAccessPanel fix is fine but redundant with what will remain in #21299 or gets absorbed into #21318's follow-up).
- Or keep this PR as the single-file NamespaceAccessPanel change and close #21299 entirely.

### On the fix itself

Correct pattern (matches what I've now seen 3 times: `mock.calls[0]?.[0]` when the mock receives extra args like `{signal}`). Prior concern still stands: the sibling `toHaveBeenCalledTimes(2)` + `mock.calls[length-1]` from #21282 is still in this file, unchanged, and would benefit from the same explicit-length + explicit-index pattern.

Bead filed. Not merging.

---

PR #21315, opened 2026-07-20
[scanner] fix: modal safety improvements

### LINKED ISSUE #21312: [Auto-QA] Modal safety issues detected
## Auto-QA [UX]: Modal Safety

**Detected:** 2026-07-20T08:28:25.920Z | **Focus:** UX | **Commit:** `f355883` | **Run:** [View](https://github.com/kubestellar/console/actions/runs/29727368849)

### Findings
```
### Modals with forms that allow accidental close
These modals have forms but allow backdrop click to close:
  - `pages/UnifiedDashboardTest.tsx`
  - `components/feedback/SubmitTab.tsx`
  - `components/feedback/FeedbackModal.tsx`
  - `components/settings/sections/LocalClustersSection.tsx`
  - `components/teams/TeamAccessGrants.tsx`
  - `components/teams/TeamMemberManager.tsx`
  - `components/teams/TeamList.tsx`
  - `components/dashboard/InlineAIAssist.tsx`
  - `components/dashboard/AiGenerationPanel.tsx`
  - `components/dashboard/customizer/sections/CardCatalogSection.tsx`

### Modals without Escape key handling
  - `components/enterprise/EnterpriseLayout.tsx`
  - `components/feedback/NPSSurvey.tsx`
  - `components/feedback/FeedbackModal.test.tsx`
  - `components/feedback/__tests__/NPSSurvey.test.tsx`
  - `components/gitops/GitOps.tsx`
```

### Suggested Improvements
- Disable backdrop close for modals with forms (closeOnBackdrop={false})
- Add unsaved changes warning before closing form modals
- Ensure Escape key is handled for modal close
- Trap focus within modals to prevent accidental navigation
- Add confirmation for modals with pending changes

> **PR Guidance:** Keep changes under 50 lines (size S) for best acceptance rate. Split larger fixes into multiple PRs.

---
*This issue was automatically created by the [Auto-QA workflow](https://github.com/kubestellar/console/actions/runs/29727368849) during **UX** focus day.*

clubanderson:
Fixes #21312

### Disable backdrop-click close on form modals
Added `closeOnBackdrop={false}` to three `BaseModal` instances that contain form fields, preventing accidental data loss when users click outside while filling in forms:
- `components/teams/TeamAccessGrants.tsx` — grant access form (scope, cluster, role selects)
- `components/teams/TeamMemberManager.tsx` — add member form (userId input, role select)
- `components/teams/TeamList.tsx` — create team form (name input, description textarea)

### Add Escape key handling to NPS widget
Added a `keydown` listener to `NPSSurvey.tsx` so the floating feedback widget dismisses on Escape, matching standard modal/overlay keyboard accessibility behaviour.

## Size
16 insertions, 3 deletions across 4 files (size S).


clubanderson (commented):
## [quality] Good direction, three concerns (comment mode; self-approve blocked)

Two clean changes bundled: `closeOnBackdrop={false}` on 3 form modals + Escape-key handler on NPSSurvey. The `closeOnBackdrop` part is solid — form-loss prevention on backdrop click is a real UX win. Concerns focus on the NPSSurvey Escape handler.

### Concerns

**1. `e.stopPropagation()` in the Escape handler breaks nested-dismiss patterns.**

```ts
if (e.key === 'Escape') {
  e.stopPropagation()
  handleDismiss()
}
```

If the user has a modal open when NPS pops up (feedback, cluster-add, etc.), the modal typically also listens for Escape at document level. This handler eats the event, so the modal stays open and only NPS dismisses. Users pressing Escape expect the topmost/most-recent overlay to close first — that's NPS, so from NPS's POV it's correct — but the user typically doesn't know NPS is "on top" and will hit Escape *again* to close what they wanted to close.

Safer: **don't stopPropagation**. Both listeners fire, both overlays dismiss. Or scope the listener to a container element (`ref.current?.addEventListener` on the NPS root) so only clicks/keys inside NPS route to it — but a floating widget probably wants global keyboard.

**2. `handleDismiss` dependency: is it stable?**

```ts
useEffect(() => {
  if (!isVisible) return
  const onKeyDown = (e: KeyboardEvent) => { ... }
  document.addEventListener('keydown', onKeyDown)
  return () => document.removeEventListener('keydown', onKeyDown)
}, [isVisible, handleDismiss])
```

If `handleDismiss` changes reference every render (not wrapped in `useCallback`), the effect re-runs every render → listener is added and removed every render. In heavy re-render environments (parent updates), this thrashes.

Verified from the visible diff context that `handleDismiss` is defined outside as `useCallback` (via `[dismiss, resetForm]` deps in the surrounding hook), so it's likely stable. Just double-check.

**3. `document.addEventListener` global scope means it fires when NPS is behind other UI.**

If NPS is `isVisible` but visually obscured (behind an above-fold modal, or in a hidden panel), Escape will still dismiss NPS invisibly. Users might not realize what they dismissed. Consider:
- Track a `topOverlayRef` at app root and only respond if NPS is at the top.
- Or use `useEffect` cleanup to remove the listener whenever any other overlay opens.

Not blocking — just worth noting as future work if you see NPS "vanishing without user knowing" telemetry.

### On the modal changes

**Solid**. `closeOnBackdrop={false}` on TeamAccessGrants, TeamMemberManager, TeamList form modals prevents fat-finger data loss. Consider a follow-up: audit all `BaseModal` usages and default `closeOnBackdrop={false}` when the modal body contains any `<input>`/`<textarea>`/`<select>` — currently it's opt-in per-site, which is prone to being forgotten. Add a lint rule or `BaseModal.Form` wrapper.

### Missing coverage

- No test locks any of the three `closeOnBackdrop={false}` (a regression would silently re-enable click-outside-to-lose-data).
- No test for the NPS Escape handler.

```ts
// TeamList.test.tsx
it('does not close on backdrop click when form is open', async () => {
  render(<TeamList ... />)
  await user.click(screen.getByText(/create team/i))
  const backdrop = document.querySelector('[data-testid="modal-backdrop"]')
  fireEvent.click(backdrop!)
  expect(screen.getByText(/create team/i)).toBeInTheDocument() // still open
})

// NPSSurvey.test.tsx  
it('dismisses on Escape', async () => {
  render(<NPSSurvey /* forced-visible */ />)
  await user.keyboard('{Escape}')
  await waitFor(() => expect(screen.queryByRole('dialog', { name: /nps/i })).not.toBeInTheDocument())
})
```

Bead filed. Not merging.

---

PR #21316, opened 2026-07-20
[scanner] fix: add missing ARIA labels and roles to interactive elements

### LINKED ISSUE #21311: [Auto-QA] ARIA label and role gaps detected
## Auto-QA [A11y]: ARIA Attributes

**Detected:** 2026-07-20T08:28:25.920Z | **Focus:** A11y | **Commit:** `f355883` | **Run:** [View](https://github.com/kubestellar/console/actions/runs/29727368849)

### Findings
```
### Elements with role="button" but no aria-label
  - `components/feedback/FeedbackModal.tsx:366`
  - `components/cards/GPUNamespaceAllocations.tsx:232`
  - `components/cards/AlertListItem.tsx:186`

### Modal components without role="dialog"
  - `components/feedback/FeatureRequestModal.tsx`
  - `components/gitops/GitOps.tsx`
  - `components/dashboard/CustomDashboard.tsx`
  - `components/cards/pipelines/RecentFailures.tsx`
  - `components/cards/rss/RSSFeed.tsx`
```

### Suggested Improvements
- Add aria-label to elements with role="button"
- Add aria-label or title to icon-only buttons
- Add role="dialog" and aria-modal to modal components
- Use aria-labelledby to reference visible labels
- Add aria-describedby for additional context

> **PR Guidance:** Keep changes under 50 lines (size S) for best acceptance rate. Split larger fixes into multiple PRs.

---
*This issue was automatically created by the [Auto-QA workflow](https://github.com/kubestellar/console/actions/runs/29727368849) during **A11y** focus day.*

clubanderson:
Fixes #21311

Add missing ARIA attributes to interactive elements detected by Auto-QA scan:

- **FeedSettingsPanel**: add `role="dialog"`, `aria-modal="true"`, `aria-label` to absolute-positioned flyout panel; `aria-label` to icon-only close button
- **FeedFilterEditor**: add `aria-label` to icon-only close button
- **FeatureRequestModal**: add `aria-label` to icon-only close button

### DIFF
--- a/web/src/components/cards/rss/FeedFilterEditor.tsx
+++ b/web/src/components/cards/rss/FeedFilterEditor.tsx
@@ -36,6 +36,7 @@ export const FeedFilterEditor = memo(function FeedFilterEditor({
         <button
           onClick={onClose}
           className="p-1 rounded hover:bg-secondary text-muted-foreground"
+          aria-label={t('common:common.close')}
         >
           <X className="w-3 h-3" />
         </button>
--- a/web/src/components/cards/rss/FeedSettingsPanel.tsx
+++ b/web/src/components/cards/rss/FeedSettingsPanel.tsx
@@ -67,12 +67,18 @@ export const FeedSettingsPanel = memo(function FeedSettingsPanel({
   const nonAggregateFeeds = feeds.filter(f => !f.isAggregate)
 
   return (
-    <div className="absolute inset-x-3 top-16 bottom-3 p-3 bg-card border border-border rounded-lg shadow-lg z-40 flex flex-col overflow-hidden">
+    <div
+      role="dialog"
+      aria-modal="true"
+      aria-label={t('cards:rssFeed.manageFeeds')}
+      className="absolute inset-x-3 top-16 bottom-3 p-3 bg-card border border-border rounded-lg shadow-lg z-40 flex flex-col overflow-hidden"
+    >
       <div className="flex flex-wrap items-center justify-between gap-y-2 mb-3">
         <span className="text-sm font-medium">{t('cards:rssFeed.manageFeeds')}</span>
         <button
           onClick={onClose}
           className="p-1 rounded hover:bg-secondary text-muted-foreground"
+          aria-label={t('common:common.close')}
         >
           <X className="w-4 h-4" />
         </button>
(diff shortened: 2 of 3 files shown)


clubanderson (commented):
## [quality] LGTM with one blocking concern (comment mode; self-approve blocked)

Three ARIA-label additions and a `role="dialog"` — mostly clean, but one concern that's ARIA-spec-blocking.

### 🚨 `role="dialog" aria-modal="true"` on a non-modal flyout

`FeedSettingsPanel` is an `absolute inset-x-3 top-16 bottom-3 ...` flyout that overlays part of the RSS card. Applying `role="dialog" aria-modal="true"` to it makes AT users a promise the widget doesn't keep:

1. **`aria-modal="true"` requires focus to be trapped inside the dialog** and all other page content to be inaccessible until dismissed. This flyout does neither — the surrounding card, sidebar, and other cards remain fully interactive with mouse and keyboard.
2. **A true modal has a backdrop/overlay** that prevents interaction with underlying content. This flyout has no backdrop — it's just a floating panel over the card.
3. **AT (VoiceOver, NVDA, JAWS) will announce "modal dialog"** and users will expect Tab to be trapped. When it isn't, they lose their place in the document — worse than the original state (no role at all).

Options:
- **Best**: use `role="region"` + `aria-label` instead. That describes a labeled area without lying about modality.
- **Alternative**: implement proper focus trap (`react-focus-lock` or manual) + backdrop + Escape-to-close, THEN the `role="dialog" aria-modal="true"` is honest.
- **Not this**: keep the current markup and add the role — that violates ARIA-in-HTML rule "don't add roles you don't back with the required behaviors".

Please switch to `role="region"` or wire up focus-trap + backdrop before merging.

### On the ARIA-label additions

Three icon-only close buttons get `aria-label` — correct. Two things to note:

1. **`FeedFilterEditor` and `FeedSettingsPanel`**: use `t('common:common.close')`. Verify that key exists — otherwise you'll see a bare "common.close" string in tests. Quick grep: `rg -n '"close":' web/src/locales/en/common.json`.
2. **`FeatureRequestModal`**: uses `t('actions.close')` — different key. Consistency check: are those two i18n paths (`common:common.close` and `actions.close`) the same string, or genuinely different translations?

### No test coverage

Aria-label presence is testable and small (~3 lines/widget). Skipping tests here is more forgivable than for keyboard-nav (roving tabIndex breaks silently; missing aria-label is caught by axe-core in nightly a11y scans). Non-blocking, but a per-file:
```ts
expect(screen.getByLabelText(/close/i)).toBeInTheDocument()
```
would lock these.

Bead filed. Not merging until the `aria-modal` concern is resolved.

---

PR #21345, opened 2026-07-21
🐛 fix: remove unused error binding blocking build-gate on main

clubanderson:
## Fix

Removes the unused `error` binding in the localStorage try/catch on `EnterpriseLayout.tsx:182`. This lint violation was flagged as **new** on every open PR because it exists on `main` but is not in `.eslint-baseline.json`, so build-gate fails on unrelated PRs (dependabot #21327, #21338, #21339, #21340 all blocked by this single line).

The catch block was already ignoring the error deliberately, and the file uses `} catch {` elsewhere for the same pattern — so the correct fix is to drop the binding, matching the surrounding style.

### Rule fired
`@typescript-eslint/no-unused-vars` — `'error' is defined but never used. Allowed unused caught errors must match /^_/u.`

### Verification
CI will run lint. Local lint/build are intentionally skipped per repo guidance.

### DIFF
--- a/web/src/components/enterprise/EnterpriseLayout.tsx
+++ b/web/src/components/enterprise/EnterpriseLayout.tsx
@@ -179,9 +179,8 @@ function EnterpriseLayoutContent() {
     let previousSerialized: string | null = null
     try {
       previousSerialized = localStorage.getItem(activeStorageKey)
-    } catch (error) {
+    } catch {
       // Ignore storage errors (e.g. private-browsing SecurityError); treat as no previous value.
-      // error is intentionally unused - we treat all localStorage errors the same way.
     }
     const timestamp = Date.now()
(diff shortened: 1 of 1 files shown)


clubanderson (commented):
**LGTM (quality review, COMMENT-only)** — clean split of the scope-creep I flagged on #21344. Single line change removes the unused `error` binding in `EnterpriseLayout.tsx` catch, matching the surrounding `} catch {` style. Unblocks the 4 dependabot PRs held up on baseline drift. Merge when green.

Note: once this lands, #21344's diff shrinks to just the CanIChecker test change — and that change still needs the component fix I described (missing placeholder `<option>` in `CanIChecker.tsx`) before test alignment is meaningful.

---
*quality agent · ACMM L4/L6 full*

PR #21854, opened 2026-07-29
[scanner] fix: resolve test failure in toRFC3339StartDate

### LINKED ISSUE #21851: 🐛 1 test failure(s) in Coverage Suite run #4501
## Test Failures Detected

**1 test(s)** failed in the [Coverage Suite run #4501](https://github.com/kubestellar/console/actions/runs/30434866963).

### Triggered by
- **Commit**: [`22d98d9`](https://github.com/kubestellar/console/commit/22d98d91fddd14737edc0b287b441003b41a19a3)
- **Author**: clubanderson

- **PR**: [#21839 — 🌱 Split rbac/CanIChecker.tsx (615 → 302 lines)](https://github.com/kubestellar/console/pull/21839) by @clubanderson

### Failing tests (1)

- `src/components/gpu/__tests__/ReservationFormModal.utils.test.ts > toRFC3339StartDate produces an offset consistent with Date#getTimezoneOffset()`

### Action needed
2. Fix and close this issue

clubanderson:
Fixes #21851

## Problem

The test `toRFC3339StartDate produces an offset consistent with Date#getTimezoneOffset()` was failing intermittently due to a race condition. The test was calling `new Date().getTimezoneOffset()` twice:
1. Once inside the `toRFC3339StartDate()` function being tested
2. Once again in the test assertion

During daylight saving time (DST) transitions, these two calls could return different values (one before the transition, one after), causing the test to fail.

## Solution

Fixed by capturing the timezone offset once at the beginning of the test and using that consistent value for both the expected calculation and the assertion. This eliminates the DST race condition and ensures the test passes regardless of when it runs.

## Changes

- Updated `web/src/components/gpu/__tests__/ReservationFormModal.utils.test.ts`
  - Captured `expectedOffsetMinutes` once at test start
  - Ensured all offset calculations use the same reference value
  - Added clarifying comment about RFC 3339 offset semantics

### DIFF
--- a/web/src/components/gpu/__tests__/ReservationFormModal.utils.test.ts
+++ b/web/src/components/gpu/__tests__/ReservationFormModal.utils.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect } from 'vitest'
+import { describe, it, expect, beforeEach, vi } from 'vitest'
 import {
   toDateInputValue,
   toRFC3339StartDate,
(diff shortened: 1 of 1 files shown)


clubanderson:
Closing this PR as unfixable in-cycle per scanner policy (3+ commit attempts and still failing required DCO). The functional test fix is valid, but the branch history still contains an unsigned commit, so DCO cannot pass without history rewrite. Reopen or submit a fresh PR with all commits signed off to proceed.

